PR #10, opened 2016-10-15
rename BitcoinSignature2016 -> EcdsaKoblitzSignature2016

harlantwood:
Request from @ChristopherA.  Would be good to get a second opinion or two before merging:  @msporny @dlongley.

### DIFF
--- a/lib/jsonld-signatures.js
+++ b/lib/jsonld-signatures.js
@@ -68,7 +68,7 @@ var libs = {};
 
 api.SECURITY_CONTEXT_URL = 'https://w3id.org/security/v1';
 api.SUPPORTED_ALGORITHMS = [
-  'BitcoinSignature2016',
+  'EcdsaKoblitzSignature2016',
   'GraphSignature2012',
   'LinkedDataSignature2015'
 ];
@@ -148,7 +148,7 @@ api.sign = function(input, options, callback) {
       JSON.stringify(api.SUPPORTED_ALGORITHMS)));
   }
 
-  if(algorithm === 'BitcoinSignature2016') {
+  if(algorithm === 'EcdsaKoblitzSignature2016') {
     if(typeof privateKeyWif !== 'string') {
       return callback(new TypeError(
         '[jsig.sign] options.privateKeyWif must be a base 58 formatted string.'));
@@ -678,7 +678,7 @@ function _verify(algorithm, input, options, callback) {
  * @param callback(err, signature) called once the operation completes.
  */
 var _createSignature = function(input, options, callback) {
-  if(options.algorithm === 'BitcoinSignature2016') {
+  if(options.algorithm === 'EcdsaKoblitzSignature2016') {
     // works same in any environment
     var signature;
     try {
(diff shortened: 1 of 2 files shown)


ChristopherA:
My motivation for not naming it after bitcoin are:

* there are other protocols that use ECDSA but with the bitcoin curve
* schnorr signatures are coming soon to bitcoin which have some favorable properties over the old elgamal derived ECDSA signatures

TLS has ciphersuites so I was thinking along those lines. The ordering for a TLS ciphersuite (i.e. TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA) is protocol, key exchange algorithm, authentication algorithm, symmetric encryption algorithm, symmetric key size, and the hashing algorithm. Notably missing is that curve used is bundled into the key exchange algorithm — we use ECDSA but with a different curve. 

To parallel TLS, we might want to add the protocol name in the front, and we may want hashing algorithm toward the end. If there are ever plans for encryption (@msporny ?), we may want to put NULL before hash algorithm. Thus maybe `lds-ecdsa_secp256k1-null-null-sha256-2016`? Or is that too much?

harlantwood:
Requesting feedback: @msporny @dlongley @auqabu

dlongley:
I think aim was rename this to `KoblitzSignature2016` instead:

https://w3c-dvcg.github.io/lds-koblitz2016/

harlantwood:
Ah yes, that link has the name as `KoblitzSignature2016`. @ChristopherA thoughts?

ChristopherA:
I'm still uncomfortable — it is Koblitz curve but an ECDSA sig (el gamal family). There will soon be spec for a Schnoor based sig, but the curve will be same.

harlantwood:
I like the specificity of your suggestions @ChristopherA, and I also appreciate the human readability of the shorter names.

If Koblitz/secp256k1 and ECDSA are the most important features of the name, what about:

`KoblitzEcdsaSignature2016`

This would be compatible with future signatures such as `KoblitzSchnoorSignature2017` and other variations.

ChristopherA:
I’m fine with this. The only other thing of significance is the hash (SHA256) but I think that is less significant.

harlantwood:
All makes sense.  So if I'm hearing well, https://w3c-dvcg.github.io/lds-koblitz2016/#examples would change to:

```js
  "signature": {
    "type": "EcdsaKoblitzSignature2017",
    "creator": "ecdsa-koblitz-pubkey:020d...",
    // ...
  }
```

Sound right?  If so, can we get some 👍  or 👎 ?   @msporny @ChristopherA @dlongley @aquabu

dlongley:
+1 to `EcdsaKoblitzSignature2016` ...it's not 2017 yet, and it's created this year :)

harlantwood:
Thanks, updated it above.

harlantwood:
Thumbs up from @msporny here as well: https://github.com/w3c-dvcg/lds-koblitz2016/issues/1#issuecomment-264999335

I've updated this PR with the new sig name, and declare it ready to merge!

PR #13, opened 2016-11-30
Infrastructure

harlantwood:
Minor tweaks; most notably adding linting to CI

- run linting and code checks in CI
- code fixes requested by jshint
- update node modules
- remove unused package and script
- add node 7 to CI
- gitignore .idea dir

PR #14, opened 2016-11-30
Add tests of promise API for Bitcoin sign & verify

harlantwood:
**Note: this diff will make more sense once #13 is merged** -- the only commit specific to this pr is 	39e052c.

Add tests of promise API for Bitcoin sign & verify.  Also, refactor to make tests independent of each other.  Note: this refactoring can and should be extended to other tests in the future.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ node_modules
 v8.log
 .c9revisions
 npm-debug.log
+.idea/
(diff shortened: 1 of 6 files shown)


harlantwood:
Anyone see any issue with this one or #13?  I'm working on more PRs that build on these.

PR #17, opened 2016-12-06
Test EcdsaKoblitzSignature2016 with security context

harlantwood:
This PR consists of 2 commits:

- 676db48 adds tests for EcdsaKoblitzSignature2016 (BTC) with a security context -- note that [test fail](https://travis-ci.org/CoMakery/jsonld-signatures/builds/181547270) after this commit, because the algorithm name contains a namespace prefix: `sec:EcdsaKoblitzSignature2016`.
- b28ba6b adds a small tweak to strip off the namespace prefix from the algorithm name.  I'm not sure why the prefix ends up there; and we only see it in the EcdsaKoblitzSignature2016 algorithm case.  Tests pass after this commit.  Any better solutions are very welcome.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ node_modules
 v8.log
 .c9revisions
 npm-debug.log
+.idea/
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,16 +2,17 @@ language: node_js
 node_js:
   - "0.10"
   - "0.12"
-  - 4
-  - 5
-  - 6
+  - "4"
+  - "5"
+  - "6"
+  - "7"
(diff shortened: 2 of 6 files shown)


dlongley:
`EcdsaKoblitzSignature2016` has been added to the live security context -- I'll need to poke around this PR a bit to adjust for that before merging.

dlongley:
Also, the tests should not even be loading the live context. It's supposed to load that local copy that's in the test file... so that will need some investigating as well.

dlongley:
Landed as d1b2cd4c92b8e3fedee5395605791d757c2afc18.

PR #19, opened 2017-03-25
fix empty message issue with EcdsaKoblitzSignature2016

kimdhamilton:
I noticed an issue with the EcdsaKoblitzSignature2016 codepaths. `_getDataToHash` returns an empty string if options.algorithm is EcdsaKoblitzSignature2016. That means it's currently signing (and verifying) the empty message.

This is a security issue for any callers using the EcdsaKoblitzSignature2016 codepath, as _any_ message given to it gets dropped. I.e. when signing, the input message is dropped and the empty message is signed. If the same signed message is tampered with, then it will also continue to verify. Again, this is because the message is being dropped during `_getDataToHash`, and the empty message will verify.

Note that I also needed to update the signatureValues in unit tests.

This PR was originally posted as https://github.com/w3c-vc/playground/pull/1, and moved here per @msporny's feedback. 

/cc: @harlantwood @dlongley @davidlehn

### DIFF
--- a/lib/jsonld-signatures.js
+++ b/lib/jsonld-signatures.js
@@ -781,7 +781,7 @@ function _getDataToHash(input, options) {
     if(options.domain !== null && options.domain !== undefined) {
       toHash += '@' + options.domain;
     }
-  } else if(options.algorithm === 'LinkedDataSignature2015') {
+  } else {
     var headers = {
       'http://purl.org/dc/elements/1.1/created': options.date,
       'https://w3id.org/security#domain': options.domain,
--- a/tests/test.js
+++ b/tests/test.js
@@ -276,6 +276,7 @@ describe('JSON-LD Signatures', function() {
       var testPublicKeyBtc;
       var testPublicKeyBtcOwner;
       var invalidPublicKeyWif;
+      var testDocumentSignedAltered;
 
       beforeEach(function() {
         testDocument = {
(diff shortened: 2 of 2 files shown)


kimdhamilton:
sorry for the noise -- left out jshint fixes

lib/jsonld-signatures.js
```diff
@@ -781,7 +781,7 @@ function _getDataToHash(input, options) {
     if(options.domain !== null && options.domain !== undefined) {
       toHash += '@' + options.domain;
     }
-  } else if(options.algorithm === 'LinkedDataSignature2015') {
+  } else if(options.algorithm === 'LinkedDataSignature2015' || options.algorithm === 'EcdsaKoblitzSignature2016') {
```
dlongley: I think we should change this to an `else` as it's the default for linked data signatures. Only `GraphSignature2012` has a special algorithm right now -- and using an `else` would future proof against this issue reoccurring with another algorithm.

kimdhamilton: That would be fine. There are already previous checks that the algorithm is supported, e.g.
```
  if(api.SUPPORTED_ALGORITHMS.indexOf(algorithm) === -1) {
    return callback(new Error(
      '[jsigs.sign] Unsupported algorithm "' + algorithm + '"; ' +
      'options.algorithm must be one of: ' +
      JSON.stringify(api.SUPPORTED_ALGORITHMS)));
  }
```

kimdhamilton:
submitted change requested by @dlongley

main.js
```diff
@@ -0,0 +1,3 @@
+/**
+ * Created by kim on 3/25/17.
+ */
```
dlongley: Did this get checked in by accident?

kimdhamilton:
done

kimdhamilton:
Please feel free to squash/rebase when merging

dlongley:
Thanks @kimdhamilton!

harlantwood:
👍👍👍

PR #26, opened 2017-09-26
Authentication credential

mattcollier:
@dlongley the test suite does not appear to use local copies of contexts.  So, public contexts will need to be updated before tests can be created for this patch.


mattcollier:
Wrong target branch.  Closing.

PR #36, opened 2018-10-03
Add proofPurpose and ppOptions properties

cwebber:
- Add proofPurpose and ppOptions fields to sign/verify procedures.
- Signing with proofPurpose supplied adds specific fields to the
  proof based on the proofPurpose's associated class's
  addFieldsToProof method.
- Verifying with proofPurpose supplied filters to only the fields
  with that proofPurpose and runs an additional check against the
  proofPurpose's verify method.
- Add LinkedDataSignature ProofPurpose as one particularly simple
  ProofPurpose suite.
- Update tests to match this expectation.

The one thing this PR doesn't do yet is update tests to test for the ProofPurpose's `addFieldsToProof` and `verify` methods.  Those would be good to test but I didn't want to hold back submitting this given deadlines.

This is the foundation for ocap-ld.js which will now be separated out.

### DIFF
--- a/lib/jsonld-signatures.js
+++ b/lib/jsonld-signatures.js
@@ -104,7 +104,7 @@ api.suites = suites;
  *            Not providing this will result in the legacy behavior
  *            of checking only signatures without a proofPurpose.  Not
  *            recommended... this can lead to confused deputy attacks.
- *          [ppOptions] Options and expectations relevant to the
+ *          [proofPurposeOptions] Options and expectations relevant to the
  *            proofPurpose object's verification method.
  *          [documentLoader(url, [callback(err, remoteDoc)])] the document
  *            loader.
@@ -175,7 +175,7 @@ api.sign = util.callbackify(async function(input, options) {
  *            an error.)  Not providing this will result in the legacy behavior
  *            of checking only signatures without a proofPurpose.  Not
  *            recommended... this can lead to confused deputy attacks.
- *          [ppOptions] Options and expectations relevant to the
+ *          [proofPurposeOptions] Options and expectations relevant to the
  *            proofPurpose object's verification method.
  * @param [callback(err, result)] called once the operation completes.
  *
--- a/lib/jsonld-signatures.js
+++ b/lib/jsonld-signatures.js
@@ -252,7 +252,7 @@ api.verify = util.callbackify(async function(input, options) {
   // of them may be is known in advance.)
   const [expanded] = await jsonld.expand(input, opts);
   const framed = await jsonld.compact(
-    input, constants.SECURITY_CONTEXT_URL, opts);
+    expanded, constants.SECURITY_CONTEXT_URL, {...opts, skipExpansion: true});
 
   // ensure there is at least one `proof` or `signature`
   const proofs = jsonld.getValues(framed, 'signature')
--- a/lib/jsonld-signatures.js
+++ b/lib/jsonld-signatures.js
@@ -53,7 +53,9 @@ const suites = {
   GraphSignature2012: require('./suites/GraphSignature2012'),
   RsaSignature2018: require('./suites/RsaSignature2018')
 };
-const proofPurposes = {};
+const proofPurposes = {
+  CredentialIssuance: require('./proof-purpose/CredentialIssuance')
+};
 
 // load locally embedded contexts
 const contexts = require('./contexts');
--- a/lib/jsonld-signatures.js
+++ b/lib/jsonld-signatures.js
@@ -100,6 +100,12 @@ api.suites = suites;
  *          [proof] a JSON-LD document with options to use for the `proof`
  *            node (e.g. `proofPurpose` or any other custom fields can be
  *            provided here using a context different from security-v2).
+ *          [proofPurpose] A ProofPurpose object.  If provided,
+ *            Not providing this will result in the legacy behavior
+ *            of checking only signatures without a proofPurpose.  Not
+ *            recommended... this can lead to confused deputy attacks.
+ *          [ppOptions] Options and expectations relevant to the
+ *            proofPurpose object's verification method.
  *          [documentLoader(url, [callback(err, remoteDoc)])] the document
  *            loader.
  * @param callback(err, signedDocument) called once the operation completes.
@@ -162,6 +168,15 @@ api.sign = util.callbackify(async function(input, options) {
  *            loader.
  *          [id] the ID (full URL) of the node to check the signature of, if
  *            the input contains multiple signed nodes.
+ *          [proofPurpose] A ProofPurpose object.  If provided, only proofs
+ *            matching this proofPurpose's URI will be checked, but will be
+ *            checked against the additional verification steps relevant to the
+ *            proofPurpose.  (Failure to find a relevant proofPurpose will throw
+ *            an error.)  Not providing this will result in the legacy behavior
+ *            of checking only signatures without a proofPurpose.  Not
+ *            recommended... this can lead to confused deputy attacks.
+ *          [ppOptions] Options and expectations relevant to the
+ *            proofPurpose object's verification method.
  * @param [callback(err, result)] called once the operation completes.
  *
  * @return a Promise that resolves to the verification result.
@@ -228,6 +243,14 @@ api.verify = util.callbackify(async function(input, options) {
   if(options.documentLoader) {
     opts.documentLoader = options.documentLoader;
   }
+  // Why expand just to frame again?  The answer is in the proofPurpose
+  // logic.  We can't anticipate all the terms we might operate on for all
+  // available ProofPurpose suites, so we expand once and let them
+  // operate on it in expanded form.
+  // (Each ProofPupose could compact to its own context, though in the case
+  // of ocap-ld, not even which caveats might be used and what the arguments
+  // of them may be is known in advance.)
+  const [expanded] = await jsonld.expand(input, opts);
   const framed = await jsonld.compact(
     input, constants.SECURITY_CONTEXT_URL, opts);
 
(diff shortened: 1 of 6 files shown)


lib/jsonld-signatures.js
```diff
@@ -228,6 +243,14 @@ api.verify = util.callbackify(async function(input, options) {
...
     opts.documentLoader = options.documentLoader;
   }
+  // Why expand just to frame again?  The answer is in the proofPurpose
+  // logic.  We can't anticipate all the terms we might operate on for all
+  // available ProofPurpose suites, so we expand once and let them
+  // operate on it in expanded form.
+  // (Each ProofPupose could compact to its own context, though in the case
+  // of ocap-ld, not even which caveats might be used and what the arguments
+  // of them may be is known in advance.)
+  const [expanded] = await jsonld.expand(input, opts);
```
dlongley: A quick comment here ... if we're expanding here and framing/compacting thereafter, pass the expanded input and the "don't expand" flag (or whatever it's called) to the framing/compacting call so we don't pay any extra price.

gannan08:
I assigned myself to this PR to resolve any and all issues.

lib/jsonld-signatures.js
```diff
@@ -100,6 +100,12 @@ api.suites = suites;
  *          [proof] a JSON-LD document with options to use for the `proof`
  *            node (e.g. `proofPurpose` or any other custom fields can be
  *            provided here using a context different from security-v2).
+ *          [proofPurpose] A ProofPurpose object.  If provided,
```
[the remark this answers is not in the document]
gannan08: ProofPurpose architecture has been refactored. Please make another pass at your earliest convenience.

gannan08:
Ready for a pass.

gannan08:
cc: @dlongley

lib/proof-purpose/CredentialIssuance.js
```diff
@@ -0,0 +1,70 @@
...
+const claimUri = credentialsNs + 'claim';
+
+module.exports = class CredentialIssuance extends ProofPurpose {
+  constructor(injector) {
+    super(injector);
+    // TODO: We need a more permanent URI for this.  Where would it go?
+    this.uri = credentialIssuanceUri;
+  }
+
+  async verify(document, proof, proofPurposeOptions) {
```
mattcollier: proofPurposeOptions is not used.  maybe this?
```suggestion
  async verify(document, proof/*, proofPurposeOptions*/) {
```

lib/proof-purpose/CredentialIssuance.js
```diff
@@ -0,0 +1,70 @@
...
+        || !document[claimUri]) {
+      return false;
+    }
+    return true;
+  }
+
+  async addFieldsToProof(proof, proofPurposeOptions) {
+    return proof;
+  }
+}
```
mattcollier: missing semi

lib/proof-purpose/CredentialIssuance.js
```diff
@@ -0,0 +1,70 @@
...
+    // fairly strange place to do this check but this is a demonstration
+    // of a highly minimal ProofPurpose class
+    if(!((document['@type'] || []).includes(VerifiableCredentialUri))
+        || !document[claimUri]) {
+      return false;
+    }
+    return true;
+  }
+
+  async addFieldsToProof(proof, proofPurposeOptions) {
```
mattcollier: Should there be a todo or a fixme or something here?  What's the purpose of this function?  Also, unused vars.

lib/proof-purpose/ProofPurpose.js
```diff
@@ -0,0 +1,60 @@
...
+// implementations are much larger than those mapped in the standard
+// ld-sigs / ocap-ld context.
+module.exports = class ProofPurpose {
+  constructor(injector) {
+    this.injector = injector;
+    // Must be overridden by child methods
+    this.uri = null;
+  }
+
+  async verify(document, proof, proofPurposeOptions) {
```
mattcollier: I suppose the params are listed in stub as documentation?
```suggestion
  async verify(*/document, proof, proofPurposeOptions*/) {
```

lib/proof-purpose/ProofPurpose.js
```diff
@@ -0,0 +1,60 @@
...
+    throw new Error(
+      '"verify" must be implemented in a derived class.');
```
mattcollier: one line

lib/proof-purpose/ProofPurpose.js
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  async addFieldsToProof(proof, proofPurposeOptions) {
```
mattcollier:
```suggestion
  async addFieldsToProof(/*proof, proofPurposeOptions*/) {
```

lib/suites/LinkedDataSignature.js
```diff
@@ -145,6 +145,14 @@ module.exports = class LinkedDataSignature {
       proof.nonce = options.nonce;
     }
 
+    // add fields from proofPurpose
+    if(options.proofPurpose) {
+      const proofPurpose = options.proofPurpose;
```
mattcollier:
```suggestion
      const {proofPurpose} = options;
```

lib/suites/LinkedDataSignature.js
```diff
@@ -145,6 +145,14 @@ module.exports = class LinkedDataSignature {
...
+      proof.proofPurpose = proofPurpose.uri;
+      await proofPurpose.addFieldsToProof(
```
mattcollier: I think it would be best to make this a conditional on `options.proofPurposeOptions`.  It is my assumption that `addFieldsToProof` would add properties to proof from `proofPurposeOptions` to the proof.  The existing implementations of this function don't do anything.

The existing implementation *does* have a return value which is not being utilized here.  So it's hard to know if it's intended to mutate `proof` or return a new proof with the options added.

lib/suites/LinkedDataSignature.js
```diff
@@ -193,6 +201,8 @@ module.exports = class LinkedDataSignature {
   async verify(framed, options) {
```
mattcollier: noting that this existing API does not conform to the preferred validation convention.

lib/suites/LinkedDataSignature.js
```diff
@@ -297,6 +307,19 @@ module.exports = class LinkedDataSignature {
         proof
       }));
 
+    // Check that the proofPurpose is valid
+    if(options.proofPurpose) {
+      const {expanded, proofPurpose} = options;
+      const verified = await proofPurpose.verify(
```
mattcollier: Should we start employing the preferred validator convention with these new `verify` APIs?  That is, proofPurpsose.verify() can return `{valid: false, error: <error>}` instead of throwing?

gannan08: Yes, we should start employing the new API.

lib/proof-purpose/CredentialIssuance.js
cwebber: While that could be done because of Javascript's goofy handling of ignoring "unused" arguments exceeding the function's actual arity, it seems like an uncomfortable choice to me given that the argument is actually passed in, and needs to be in the general case for these classes.

lib/proof-purpose/ProofPurpose.js
cwebber: See my comment on arity previously.  I think the desire to comment out arguments when we're specifying what the API will use seems very strange.  Is this to prevent the linter from complaining, or something?

lib/proof-purpose/CredentialIssuance.js
mattcollier: is the `addFieldsToProof` API designed to mutate `proof`?

lib/suites/LinkedDataSignature.js
mattcollier: @cwebber is there a reason to call addFieldsToProof if options `proofPurposeOptions` is not defined?

cwebber: > I think it would be best to make this a conditional on options.proofPurposeOptions. It is my assumption that addFieldsToProof would add properties to proof from proofPurposeOptions to the proof.

I guess that's probably true, but maybe not necessarily.  I'm struggling to think of an example that wouldn't be gathered from proofPurposeOptions; it could be that such an example would be something dynamic from the environment such as something time related, though I'll admit that dynamic elements are uncomfortable to ponder.

However, even the absence of proofPurposeOptions could warrant calling this function, as the ProofPurpose class may raise an exception at the *absence* of an option.

> The existing implementations of this function don't do anything.

I guess I hadn't gotten to it, but the intention was that ocap-ld.js would use it to add the`capability` field to the proof.  @gannan08 take note for ocap-ld.js :)

>  The existing implementation does have a return value which is not being utilized here. So it's hard to know if it's intended to mutate proof or return a new proof with the options added.

Yes, we should make a decision on mutate or return a new replacement.

lib/jsonld-signatures.js
```diff
@@ -126,11 +137,21 @@ api.sign = util.callbackify(async function(input, options) {
 
   options = _addEmbeddedContextDocumentLoader(options);
 
+  let {proofPurpose} = options;
+  if(proofPurpose) {
+    // FIXME: check if registered
```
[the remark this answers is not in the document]
cwebber: I didn't write this (@gannan08 added the registry) but I'd think line 144 would catch it?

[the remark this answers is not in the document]
gannan08: Yes, the FIXME has been addressed on [L144](https://github.com/digitalbazaar/jsonld-signatures/pull/36/files/b4bdc16ae3c49fdf9d0c74cd692da7346359b34b#diff-a1df416cb75df6428c21b0be5fd80886R144)
@cwebber @mattcollier

lib/jsonld-signatures.js
```diff
@@ -240,12 +280,37 @@ api.verify = util.callbackify(async function(input, options) {
...
+    }
+    proofPurpose = new ProofPurpose(injector);
+  }
+  const filteredProofs = proofs.filter(
+    ({doc}) => {
+      const proof = doc;
+      if(proofPurpose) {
+        return proof.proofPurpose === proofPurpose.uri;
+      } else {
+        return !proof.proofPurpose;
```
dlongley: Remove unnecessary `else` everywhere a return occurs in related `if`.

PR #37, opened 2018-11-13
Proof purpose

mattcollier:


lib/constants.js
```diff
@@ -4,6 +4,7 @@
 'use strict';
 
 module.exports = {
+  CREDENTIALS_CONTEXT_URL: 'https://w3id.org/credentials',
```
dlongley: This and the related `CredentialIssuance` proof purpose should not be added to this core lib. It should be provided via something like vc.js.

dlongley:
Note that `proofPurposeOptions` in `verify` also needs to take in accepted cryptosuites:

https://github.com/w3c/vc-data-model/issues/275#issuecomment-438407145

lib/constants.js
mattcollier: Setting aside CredentialIssuance.js here for now: https://gist.github.com/mattcollier/8a2a0e7cb80965c45e03ad8b0cd32fc8

mattcollier:
@dlongley @gannan08 the `purpose` option given to proofPurpose handlers looks like this:
```
{
  proofPurpose: 'foo'
}
```
The entire purpose object is passed into the handlers, so handlers can effectively define their own options.  Sound good?

Regarding cryptoSuites, based on your comments we need to have terms defined in contexts for at least two cryptoSuite values for every proofPurpose we intend to use?

RsaSignatureCredentialIssuance2018, Ed25519SignatureCredentialIssuance2018
RsaSignatureSomeOtherCapability2018, Ed25519SignatureSomeOtherCapability2018

Is this right?  Why do we need this proliferation of cryptoSuites?

I captured this DID document from recent code: https://gist.github.com/mattcollier/123bcd73881e9160cb522bca857be673

Are all the types on keys correct there?

lib/jsonld-signatures.js
```diff
@@ -228,24 +264,61 @@ api.verify = util.callbackify(async function(input, options) {
   if(options.documentLoader) {
     opts.documentLoader = options.documentLoader;
   }
+  // Why expand just to frame again?  The answer is in the proofPurpose
```
mattcollier: Should these comments be left in?

lib/jsonld-signatures.js
```diff
@@ -263,12 +336,12 @@ api.verify = util.callbackify(async function(input, options) {
       //const Suite = require('./suites/' + algorithm);
       const Suite = suites[algorithm];
       const verified = await new Suite(injector).verify(
-        f, Object.assign({}, options, {framed}));
+        f, Object.assign({}, options, {expanded, framed, purpose}));
```
mattcollier: Do we still need `expanded` at this point?  Relates to the comments above on L268.

lib/suites/LinkedDataSignature.js
```diff
@@ -297,9 +310,21 @@ module.exports = class LinkedDataSignature {
         proof
       }));
 
+    // Check that the proofPurpose is valid
+    const {purpose} = options;
+    if(purpose) {
+      const {expanded, purpose: {proofPurposeHandler}} = options;
+      const result = await proofPurposeHandler.verify(
```
mattcollier: related to other comments, we do want to pass proofs to verify in expanded form?

lib/suites/LinkedDataSignature.js
```diff
@@ -193,12 +204,14 @@ module.exports = class LinkedDataSignature {
   async verify(framed, options) {
     options = Object.assign({}, options || {});
 
+    // TODO: This only supports one proof object at a time.
```
mattcollier: I believe this comment can be remove, multiple proofs are properly handled in jsonld-signatures.js.  Each `LinkedDataSignature` is verified one at a time at this level.

dlongley:
@mattcollier,

> The entire purpose object is passed into the handlers, so handlers can effectively define their own options. Sound good?

Yes.

> Regarding cryptoSuites, based on your comments we need to have terms defined in contexts for at least two cryptoSuite values for every proofPurpose we intend to use?
> 
> RsaSignatureCredentialIssuance2018, Ed25519SignatureCredentialIssuance2018
> RsaSignatureSomeOtherCapability2018, Ed25519SignatureSomeOtherCapability2018
> 
> Is this right? Why do we need this proliferation of cryptoSuites?

I would also prefer we not have that many... and that we come up with a better name than "crypto suites". So, from the CCG calls, what I recall is that the type for these things is meant to convey:

1. The type of parameters one requires in order to verify a proof.
2. A certain level of security and/or assurance and a relative "freshness" (via a year).

An instance of one of these things contains the actual parameter values. So, for example, an `RsaSignatureCredentialIssuance2018` indicates that there is just one parameter, `publicKey`, whose value must be an `RsaVerificationKey2018`.

Perhaps we should be calling these "verification suites" instead of "crypto suites". Another example for one of these that doesn't use a cryptographic key would be:

`FingerprintVerification2018` which would indicate there are parameters, `fingerprintTemplate` and `fingerprintVerifyService`, specifying some kind of salted, pseudonymous template that must be used in conjunction with a finger print verification service and the data privately collected from the fingerprint itself, in order to verify someone. Or something like that.

So, if I've got this right, it seems to me that we could simplify these such that there would be only one type per type of key. For example,

`RsaSignatureVerification2018` which plays the same role that `RsaSignatureCredentialIssuance2018`, `RsaSignatureCapabilityDelegation2018`, `RsaSignatureCapabilityInvocation2018`, and so on... would play. I think we have already captured what the "purpose" of using these verification suites is via the relationship between the entity (e.g. expressed in DID document) and the suites. In other words, if you follow the `capabilityInvocation` property to its value, you will find a verification suite.

So I don't think there's additional value in creating a separate suite for each one of these ... but rather, what should govern whether or not we need an additional suite is the type of verification (e.g. biometric, digital signature, so on), its security/assurance/freshness, and what parameters are needed.

For example, if you were trying to verify a capability invocation proof, you would specify to a verifier tool which types of verification suites you support (e.g. RSA or Ed25519 signature verification). The verifier tool would search for a verification suite that matched underneath the `proofPurpose` in the key controller's document. If one of those matching verification suites held the key used in the proof as a parameter, then it would be considered acceptable (of course, the signature itself needs to be checked and so on). The same would apply if you had specified that you accept a fingerprint-based proof.

Thoughts? @mattcollier, @msporny, @dmitrizagidulin, @gannan08 ?

lib/jsonld-signatures.js
dlongley: They seem like they are trying to be helpful but aren't quite there. So I'd like to keep them in, but have them actually say what "the answer" is rather than guessing what the author's intent was. For example:

```Here we expand and then frame again because ...```

Also... there's a flag to skip expansion when framing and doing other ops if the input is known to already be expanded. I didn't look to see if we're doing that here ... just responding to the comment that only exposed so much contextual code.

**Comment by dmitrizagidulin** [comment]
@dlongley +1 for calling these "verification suites" instead of "crypto suites"!

And I like the idea of simplifying proof purposes to a handful of `*SignatureVerification2018`, one per key type.

**Comment by mattcollier** [comment]
@gannan08 are you able to assess whether this API suits your needs in ocapld.js?

**Comment by mattcollier** [comment]
@gannan08 @dlongley  I got to thinking about how proof purposes are added using this API and I don't think I like it.

https://github.com/digitalbazaar/ocapld.js/blob/refactor/lib/index.js#L18

I think exposing `jsigs.proofPurposes` is not good.  Mainly because we already have an established injector pattern in this module which is being usurped here.

So I got to thinking about using the already existing injector `jsigs.use`.  Although technically, this would work fine, it seems weird to me to combine multiple classes of things in the same injector.  So the main injector would include `jsonld` and `FooProofPurpose` which seems odd.

So, instead I'm thinking we could do `jsigs.proofPurpose.use` whereby `jsigs.proofPurpose` would have a separate injector instance to track proofPurpose handlers.

What do you think about that?

**Comment by dlongley** [comment]
@mattcollier,

> What do you think about that?

I'm ok with it ... we should do the same with suites to be consistent, but we might need some backwards compatibility stuff there to enable the old style before deprecating it.

File: lib/jsonld-signatures.js (hunk 2 above)
Comment (mattcollier) [reply lib/jsonld-signatures.js]: @dlongley a couple points of clarification.  The statements made in the comments are accurate wrt the necessity of working with expanded data?

Second, the prose and variable names are referring to `framed` but the operation being referred to is `jsonld.compact` (which is btw using the skipExansion flag).

https://github.com/digitalbazaar/jsonld-signatures/blob/4d8384ceac5082da0b0c2cb432ddcaca295ac23a/lib/jsonld-signatures.js#L267-L279

Comment (dlongley) [reply lib/jsonld-signatures.js]: @mattcollier, I'm pretty sure @cwebber wrote those statements... @cwebber can you weigh in on what's going on here? @gannan08 you may also have comments.

Comment (mattcollier) [reply lib/jsonld-signatures.js]: @dlongley upon closer inspection the use of the variable `framed` for a `jsonld.compact` operation predates this PR: https://github.com/digitalbazaar/jsonld-signatures/pull/37/files#diff-a1df416cb75df6428c21b0be5fd80886R275

I'm just expecting that `framed` would be associated with a `frame` operation.  Maybe this is not an accurate expectation.

**Comment by mattcollier** [comment]
@dlongley @gannan08 proofPurposes are injectedt like this:
```
jsigs.proofPurposes.use('NoOpProofPurpose', NoOpProofPurpose);
```
https://github.com/digitalbazaar/jsonld-signatures/blob/3e14fbc7d875ca43e72c560e6e5f33ccfd84dc25/tests/test-common.js#L93-L94

gannan08:
@mattcollier 
> @dlongley @gannan08 the `purpose` option given to proofPurpose handlers looks like this:
> 
> ```
> {
>   proofPurpose: 'foo'
> }
> ```
> The entire purpose object is passed into the handlers, so handlers can effectively define their own options. Sound good?

Looks good.
> 
> Regarding cryptoSuites, based on your comments we need to have terms defined in contexts for at least two cryptoSuite values for every proofPurpose we intend to use?
> 
> RsaSignatureCredentialIssuance2018, Ed25519SignatureCredentialIssuance2018
> RsaSignatureSomeOtherCapability2018, Ed25519SignatureSomeOtherCapability2018
> 
> Is this right? Why do we need this proliferation of cryptoSuites?
> 
> I captured this DID document from recent code: https://gist.github.com/mattcollier/123bcd73881e9160cb522bca857be673
> 
> Are all the types on keys correct there?

All but one of the key types are correct. The type for the authentication key should be `Ed25519SignatureAuthentication2018`.

lib/jsonld-signatures.js
gannan08: @mattcollier @dlongley We no longer need to pass the expanded input into proofPurpose handlers. Documents passed into the proofPurpose handlers and documents that are outputted are expected to be compacted using the security context. Expansion should be handled in the proofPurpose libs when necessary.

mattcollier:
@gannan08 I would like to assign this (back) to you to complete since you know how the inputs to proof purpose handlers need to be crafted.  All the outstanding comments that I'm aware of are related to that.

gannan08:
> @gannan08 are you able to assess whether this API suits your needs in ocapld.js?

I'm currently integrating with API right now and everything seems to be fine at the moment.

gannan08:
@mattcollier 

> @gannan08 I would like to assign this (back) to you to complete since you know how the inputs to proof purpose handlers need to be crafted. All the outstanding comments that I'm aware of are related to that.

Okay great, sounds good.

PR #38, opened 2018-11-15
Implement signer API.

mattcollier:
@dlongley @dmitrizagidulin does this accurately reflect what we discussed?

@davidlehn do you know anything about the FIXME related to babel and Buffer?

### DIFF
--- a/lib/suites/Ed25519Signature2018.js
+++ b/lib/suites/Ed25519Signature2018.js
@@ -40,7 +40,21 @@ module.exports = class Ed25519Signature2018 extends LinkedDataSignature {
       JSON.stringify(header), {forge});
 
     let encodedSignature;
-    if(this.injector.env.nodejs) {
+    const {nodejs} = this.injector.env;
+    if(options.signer) {
+      // a custom signing API has been provided
+      // a Uint8Array is passed to the signer
+      const {data} = util.createJws({encodedHeader, forge, nodejs, verifyData});
+      let message;
+      if(nodejs) {
+        // convert Buffer to Uint8Array
+        message = new Uint8Array(data.buffer);
+      } else {
+        // convert binary encoded string to Uint8Array
+        message = forge.util.binary.raw.decode(data);
+      }
+      encodedSignature = await options.signer.sign({message});
+    } else if(nodejs) {
       // optimize using node libraries
       const chloride = require('chloride');
       const bs58 = require('bs58');
@@ -49,10 +63,9 @@ module.exports = class Ed25519Signature2018 extends LinkedDataSignature {
       const privateKey = bs58.decode(options.privateKeyBase58);
 
       // build signing input per above comment
-      const tbs = Buffer.concat([
-        Buffer.from(encodedHeader + '.', 'utf8'),
-        Buffer.from(verifyData.data, verifyData.encoding)]);
-      const buffer = chloride.crypto_sign_detached(tbs, privateKey);
+      const {data} = util.createJws(
+        {encodedHeader, forge, nodejs, verifyData});
+      const buffer = chloride.crypto_sign_detached(data, privateKey);
       encodedSignature = util.encodeBase64Url(
         buffer.toString('binary'), {forge});
     } else {
@@ -61,14 +74,10 @@ module.exports = class Ed25519Signature2018 extends LinkedDataSignature {
       const privateKey = forge.util.binary.base58.decode(
         options.privateKeyBase58);
       // build signing input per above comment
-      const buffer = new forge.util.ByteBuffer(encodedHeader + '.', 'utf8');
-      buffer.putBuffer(new forge.util.ByteBuffer(
-        verifyData.data, verifyData.encoding));
-        const binaryString = forge.util.binary.raw.encode(
-          forge.ed25519.sign({
-            message: buffer,
-            privateKey
-          }));
+      const message = util.createJws(
+        {encodedHeader, forge, nodejs, verifyData});
+      const binaryString = forge.util.binary.raw.encode(
+        forge.ed25519.sign({message, privateKey}));
       encodedSignature = util.encodeBase64Url(binaryString, {forge});
     }
 
--- a/lib/suites/Ed25519Signature2018.js
+++ b/lib/suites/Ed25519Signature2018.js
@@ -12,6 +12,17 @@ module.exports = class Ed25519Signature2018 extends LinkedDataSignature {
     this.requiredKeyType = 'Ed25519VerificationKey2018';
   }
 
+  /**
+   * @param verifyData {object}
+   *
+   * @param options {object} Contains key material, either a `signer` object
+   *   or a base-58 encoded private key.
+   *
+   * @param [options.privateKeyBase58] {string}
+   * @param [options.signer] {object}
+   *
+   * @returns {Promise<string>} Signature as an encoded detached JWS
+   */
   async createSignatureValue(verifyData, options) {
     const forge = this.injector.use('forge');
 
--- a/lib/suites/Ed25519Signature2018.js
+++ b/lib/suites/Ed25519Signature2018.js
@@ -7,23 +7,28 @@ const LinkedDataSignature = require('./LinkedDataSignature');
 const util = require('../util');
 
 module.exports = class Ed25519Signature2018 extends LinkedDataSignature {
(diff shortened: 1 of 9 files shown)


lib/suites/Ed25519Signature2018.js
```diff
@@ -73,6 +75,19 @@ module.exports = class Ed25519Signature2018 extends LinkedDataSignature {
 
     // create detached content signature
     return encodedHeader + '..' + encodedSignature;
+
+    function _createSigningString() {
+
+      // FIXME: we need a Buffer implementation that works for browser
```
dmitrizagidulin: Yes, Webpack provides a buffer polyfill (though from what I understand, @dlongley doesn't like it? not quite sure why.)

lib/suites/Ed25519Signature2018.js
```diff
@@ -40,7 +40,11 @@ module.exports = class Ed25519Signature2018 extends LinkedDataSignature {
       JSON.stringify(header), {forge});
 
     let encodedSignature;
-    if(this.injector.env.nodejs) {
+    if(options.signer) {
+      // a custom signing API has been provided
+      const signingString = _createSigningString();
+      encodedSignature = await options.signer.api({signingString});
```
dmitrizagidulin: Should this be `signer.sign()`? (instead of `signer.api()`?)

`.api()` reads like it returns a whole API, whereas I assume it's an invocation to sign something?

lib/suites/Ed25519Signature2018.js
dlongley: In general, we should be writing code for the browser and polyfilling node, not the other way around. This is especially true when some core feature is currently never expected to be implemented in the browser. One property of polyfills is that they are intended to be temporary; they'll eventually go away, removing the need for sending them to the browser. In this light, the `Buffer` "polyfill" really isn't a polyfill. The `Buffer` "polyfill" is a foreverfill.

That being said, I'm not opposed to compromise in the interest of time. But we shouldn't be treating this sort of problem the same as, for example, using async/await or Promises and having to polyfill those for some browsers. This isn't write once and wait for browsers to catch up and we get to remove the overhead. This overhead will stick around until we manually remove it -- and that sort of overhead tends to pile up and overwhelm you before you know it.

lib/Injector.js
```diff
@@ -12,6 +12,7 @@ module.exports = class Injector {
     if(!this.env.nodejs) {
       this.env.browser = true;
     }
+    console.log('*************DETECTED ENVIRONMENT********', this.env);
```
davidlehn: Remove debugging.

lib/suites/Ed25519Signature2018.js
mattcollier: @dlongley another compromise here would be to pass the `signingString` to the handler in hex or base64, and let the handler implement a buffer if required in whatever way it chooses.  (e.g. forge)

dlongley: Also, looking at this, why do we need to use a Buffer here at all? Or, alternatively, check `verifyData.encoding` and use a Buffer when it makes sense, otherwise just do string concatenation for a `binary` string encoding -- which is probably what the browser version is doing.

mattcollier: When we discussed this API the other day we said that it would accept a string or buffer.  I'm fine with always passing in a binary encoded string.  That *is* what you are suggesting?

lib/suites/Ed25519Signature2018.js
mattcollier: I wonder if `signer.api.sign` would not be the thing?

dmitrizagidulin: `signer.api.sign()` seems reasonable; my only question then is - are there any other functions in `.api`, other than sign?

lib/suites/Ed25519Signature2018.js
mattcollier: We've decided to use unit8array.

dmitrizagidulin:
👍 from me on this PR

lib/suites/Ed25519Signature2018.js
```diff
@@ -40,7 +40,21 @@ module.exports = class Ed25519Signature2018 extends LinkedDataSignature {
...
+      const {data} = _createSigningString({nodejs});
+      let bytesUint8Array;
+      if(nodejs) {
+        // convert Buffer to Uint8Array
+        bytesUint8Array = new Uint8Array(data.buffer);
+      } else {
+        // convert binary encoded string to Uint8Array
+        bytesUint8Array = forge.util.binary.raw.decode(data);
+      }
+      encodedSignature = await options.signer.sign({bytesUint8Array});
```
dlongley: Recommend we name it `data` or `message` instead of `bytesUint8Array` as a more common sign API parameter name.

lib/suites/LinkedDataSignature.js
```diff
@@ -89,6 +89,19 @@ module.exports = class LinkedDataSignature {
     if(options.nonce !== undefined && typeof options.nonce !== 'string') {
       throw new TypeError('"options.nonce" must be a string.');
     }
+    const {signer} = options;
+    if(signer) {
+      if(!(signer.sign && typeof signer.sign === 'function')) {
+        throw new TypeError('"options.signer.sign" must be a function.');
+      }
+      if(!(signer.creator && typeof signer.creator)) {
```
dlongley: I'm wondering if `creator` should be passed to the sign function instead? I guess someone crafting the `sign` function could do this on their own ... any thoughts here? Seems natural to me that the `sign` function would receive the ID of the key to be used and the data to be signed ... and it would return an object that includes at least the raw signature (we may want to allow for it to include other meta data in the future, I don't know).

lib/suites/Ed25519Signature2018.js
mattcollier: using `message`. Done.

mattcollier:
@dmitrizagidulin this is ready for you to use in your projects.  We'll wait to merge this until you provide positive feedback based on your implementation experience.

dmitrizagidulin:
@mattcollier ok, excellent, will do

dmitrizagidulin:
@mattcollier Can we merge this PR into the `proofPurpose` branch? I don't want those diverging too much..

dmitrizagidulin:
@mattcollier Thanks! :)

mattcollier:
@dmitrizagidulin I just rebased onto @gannan08 proof-purpose-test branch, tests are green, and I think he's wrapping up.

dmitrizagidulin:
@mattcollier @dlongley ok, so, I think we need a slightly different level of abstraction here. 

Instead of invoking signer.sign() inside each suite (so, currently it's invoked in Ed25519Signature2018.js, and in RsaSignature2018.js), it should be done one layer up, in `LinkedDataSignatures.js`. 

So currently, [LinkedDataSignatures.createProofNode](https://github.com/digitalbazaar/jsonld-signatures/blob/8e9841b5714fb149e31aa0e3b617621ea9b5c73e/lib/suites/LinkedDataSignature.js#L185) is:

```js
  async createProofNode(verifyData, options) {
    const proof = options.proof;
    proof.jws = await this.createSignatureValue(verifyData, options);
    return proof;
  }
```

And instead it should be something like:

```js
  async createProofNode(verifyData, options) {
    const proof = options.proof;
    if(options.signer) {
      proof.jws = await signer.sign(verifyData, options);
    } else {
      proof.jws = await this.createSignatureValue(verifyData, options);
    }
    
    return proof;
  }
```

I think that matches the intention for having a general sign() override option?

mattcollier:
@dmitrizagidulin no objection here, especially if there is a net reduction in the amount of code.

mattcollier:
@dmitrizagidulin please make it do what you need.

dmitrizagidulin:
@mattcollier ok, cool. Doing so..

tests/test-common.js
```diff
@@ -1133,6 +1181,40 @@ describe('JSON-LD Signatures', function() {
         });
       });
 
+      it.skip('signs a local document using a custom signer', async () => {
```
dlongley: This and the test above are skipped... why? Can we enable them?

dmitrizagidulin: Oh whoops, I forgot about these. They're disabled cause I changed the API, these no longer apply. But I meant to add equivalent tests and remove these. Thanks!

PR #39, opened 2018-11-29
Proof purpose updates

gannan08:


gannan08:
@davidlehn @dlongley @mattcollier  Ready for a review, all tests pass. Please take an extra look at the updates: changes to context, renaming of variables, and refactor of public APIs.

lib/jsonld-signatures.js
```diff
@@ -274,6 +274,7 @@ api.verify = util.callbackify(async function(input, options) {
   // of ocap-ld, not even which caveats might be used and what the arguments
   // of them may be is known in advance.)
   const [expanded] = await jsonld.expand(input, opts);
+
```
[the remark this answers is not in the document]
gannan08: I was going to address that issue once I merged this branch into [#proofPurpose](https://github.com/digitalbazaar/jsonld-signatures/tree/proofPurpose).

lib/jsonld-signatures.js
```diff
@@ -364,6 +371,20 @@ function _getSupportedAlgorithms() {
   return Object.keys(api.suites).filter(s => s !== 'LinkedDataSignature');
 }
 
+function _getSupportedProofPurposes() {
```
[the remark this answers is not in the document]
gannan08: Correct, fixing that now.

lib/contexts/security-v2.js
```diff
@@ -7,10 +7,17 @@ module.exports = {
     "EquihashProof2018": "sec:EquihashProof2018",
     "RsaSignature2018": "sec:RsaSignature2018",
     "RsaVerificationKey2018": "sec:RsaVerificationKey2018",
-
+    "capability": {"@id": "sec:capability", "@type": "@id"},
```
dlongley: Seems like we're missing `allowedAction` and `capabilityAction` terms like are in the new DID context:

https://github.com/w3c-ccg/did-spec/blob/gh-pages/did-v0.11.jsonld#L19

lib/jsonld-signatures.js
```diff
@@ -141,23 +141,23 @@ api.sign = util.callbackify(async function(input, options) {
...
   }
   // TODO: won't work with static analysis?
   // use signature suite
   //const Suite = require('./suites/' + algorithm);
   const Suite = suites[algorithm];
-  return new Suite(injector).sign(input, options);
+  return new Suite(injector).sign(
+    input,
+    Object.assign(
+      {}, options, {purpose, proofPurposeHandler, purposeParameters}));
```
dlongley: If this does the same thing as `Object.assign()` ... do it:

`{...options, purpose, proofPurposeHandler, purposeParameters}`

lib/jsonld-signatures.js
```diff
@@ -283,35 +284,40 @@ api.verify = util.callbackify(async function(input, options) {
...
-    if(!ProofPurpose) {
-      throw new Error(`Unsupported proof purpose handler "${proofPurpose}".`);
-    }
-    purpose.proofPurposeHandler = new ProofPurpose(injector);
+  const {purpose, purposeParameters} = options;
+  const SUPPORTED_PROOF_PURPOSES = _getSupportedProofPurposes();
+  if(SUPPORTED_PROOF_PURPOSES.indexOf(purpose) === -1) {
+    // TODO: Should we throw an error here? This will break all usages of
+    // json-ld signatures. Is it true that a proofPurpose is required for
+    // all signatures?
```
dlongley: Yes, we should require it. All signature proofs should have a purpose associated with them.

dlongley: If we need one iteration of a backwards compatible version, we can insert a default proof purpose that just checks against `publicKey` (not any special verification suite). But a major point release should remove this default behavior because it's dangerous.

lib/jsonld-signatures.js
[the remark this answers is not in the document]
gannan08: Went ahead and ripped it out.

lib/jsonld-signatures.js
gannan08: Using the spread operator causes karma tests to fail, I can add the object spread operator polyfill to allow for what you suggested above.

gannan08:
Ready for another pass.

lib/jsonld-signatures.js
```diff
@@ -364,6 +357,32 @@ function _getSupportedAlgorithms() {
   return Object.keys(api.suites).filter(s => s !== 'LinkedDataSignature');
 }
 
+function _isSupportedProofPurpose(purpose) {
+  if(!purpose) {
+    // TODO: Enable error in the next major release
```
mattcollier: We should connect this with an issue, otherwise it will linger here unnoticed.

lib/jsonld-signatures.js
```diff
@@ -364,6 +357,32 @@ function _getSupportedAlgorithms() {
...
+    // throw new Error(`Please provide 'purpose' of the proof to verify.`);
+    return;
+  }
+  try {
+    api.proofPurposes.use(purpose);
```
mattcollier: The `injector.use` API does not throw when an unknown handler is specified, it just returns undefined.

lib/jsonld-signatures.js
```diff
@@ -364,6 +357,32 @@ function _getSupportedAlgorithms() {
...
+}
+
+async function _getExpandedProofPurpose({proof, jsonldOpts}) {
+  const {proofPurpose} = proof;
+  const jsonld = injector.use('jsonld');
+
+  const expanded = await jsonld.expand({
+    proofPurpose,
+    '@context': constants.SECURITY_CONTEXT_URL}, jsonldOpts);
+  return expanded[0]['https://w3id.org/security#proofPurpose'][0]['@id'];
```
mattcollier: Everything has been validated up to this point to ensure that this is not going to fail, correct?

lib/suites/Ed25519Signature2018.js
```diff
@@ -64,10 +64,11 @@ module.exports = class Ed25519Signature2018 extends LinkedDataSignature {
       const buffer = new forge.util.ByteBuffer(encodedHeader + '.', 'utf8');
       buffer.putBuffer(new forge.util.ByteBuffer(
         verifyData.data, verifyData.encoding));
-      const binaryString = forge.ed25519.sign({
-        message: buffer,
-        privateKey
-      }).toString('binary');
+        const binaryString = forge.util.binary.raw.encode(
```
mattcollier: good catch on this!

lib/jsonld-signatures.js
gannan08: It didn't return undefined when I used it. Here's a test I created that passes an unknown proof purpose https://github.com/digitalbazaar/jsonld-signatures/pull/39/commits/1614348db531fbe67bbe253df27cb5e117297c61

mattcollier: Here's the injector, it does not throw: https://github.com/digitalbazaar/jsonld-signatures/blob/master/lib/Injector.js

mattcollier: and if you are seeing an error here, what is it?

lib/jsonld-signatures.js
gannan08: If by validation, you mean code to check that this structure exists... the answer is no. I'm assuming that this structure is guaranteed by the use of the security context and the term definition of `proofPurpose` in that context. May need @dlongley @msporny @davidlehn to verify my JSON-LD logic.

lib/jsonld-signatures.js
gannan08: The error is thrown after executing [Line 45](https://github.com/digitalbazaar/jsonld-signatures/blob/master/lib/Injector.js#L45). It's due to `require('UnknownProofPurpose')`, which also brings up another issue. If we passed in a proofPurpose that we didn't register AND it matches a name of a library installed in `node_modules/`, the current behavior of the Injector will require the already existing lib. This will definitely cause some unexpected errors.

mattcollier: hmmm... yeah, we don't want any of that loading of default business for the proofPurpose.

maybe we need a kill switch option to the injector like `useDefaults: false` when using it in the proof purpose context.

gannan08: Should we name the parameter `useDefault` or `useDefaults`? I like the singular version.

mattcollier: +1 to `useDefault` as I prepare to lose another singular vs plural argument.

lib/jsonld-signatures.js
mattcollier: todo pending results of conversation in issue: https://github.com/digitalbazaar/jsonld-signatures/issues/41

lib/jsonld-signatures.js
```diff
@@ -266,52 +265,43 @@ api.verify = util.callbackify(async function(input, options) {
...
-    }
-    const ProofPurpose = api.proofPurposes.use(proofPurpose);
-    if(!ProofPurpose) {
-      throw new Error(`Unsupported proof purpose handler "${proofPurpose}".`);
-    }
-    purpose.proofPurposeHandler = new ProofPurpose(injector);
-  }
+  const {purpose, purposeParameters} = options;
+
+  _isSupportedProofPurpose(purpose);
```
mattcollier: now that I understand better what's going on here... :)

I think this needs to be moved up and maybe the todo in the helper should be here about throwing in future versions.

We're doing a lot of work before this point just to potentially throw a SyntaxError (missing purpose parameter)

lib/jsonld-signatures.js
dlongley: What if `proofPurpose` is undefined? What if `proofPurpose` does not map to `https://w3id.org/security#proofPurpose`? If there aren't other logical guarantees to ensure these things are true before the above code is run, it could blow up. We should have tests.

lib/jsonld-signatures.js
gannan08: > now that I understand better what's going on here... :)
> 
Awesome!
> I think this needs to be moved up and maybe the todo in the helper should be here about throwing in future versions.
> 
> We're doing a lot of work before this point just to potentially throw a SyntaxError (missing purpose parameter)

Okay will do, fix is here: https://github.com/digitalbazaar/jsonld-signatures/pull/39/commits/6ffaf11236bcfa283bd2e1a7b187145340ad9b81

mattcollier:
@gannan08 since we have another branch that has just been rebased on this, please do not do any squashing etc on this branch since there were some merge conflicts.  If we need to do any squashing, we can do that after this branch is merged to proofPurpose.

PR #47, opened 2018-12-23
Update dependencies.

davidlehn:
@dlongley Not sure how ready this is to merge and release.  It helps with some issues, but also has new ones.  The non-min output now uses the "eval" style so is difficult to debug and is very large.  It's pulling in lots of code (forge, jsonld, core-js) so even minimized output is >400k.  The .min.js.map file is also large.  Thoughts?

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -48,6 +48,7 @@ module.exports = function(config) {
     },
 
     webpack: {
+      mode: 'production',
       devtool: 'inline-source-map',
       module: {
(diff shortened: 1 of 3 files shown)


dlongley:
This needs a rebase ... and it seems to me that we shouldn't be "including" jsonld-signatures or rdf-canonize as this is a library with dependencies on those things, not an application bundle. The dependency updates would be good to have once rebased though.

PR #50, opened 2019-01-04
Use external crypto-ld library for LDKeyPair class.

dmitrizagidulin:
Uses the extracted `LDKeyPair` class from `crypto-ld`. WIP:
* Mock keys are failing the new parameter validation (they are bit length 1024), so they need to be re-generated
* Mocha tests pass (if you disable param validation)
* Karma tests are failing because it's not excluding `ursa` and `crypto` for some reason. What am I missing, webpack or karma config wise?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # jsonld-signatures ChangeLog
 
+### Changed
+- Use crypto-ld@2.
+
 ### Removed
 - **BREAKING**: Remove Node.js 6.x support. If you need Node.js 6.x support
   please use the 2.x series or setup your own translation.


dlongley:
@dmitrizagidulin -- please rebase this, I updated the test suite to fix the 2048-bit RSA key issue.

dmitrizagidulin:
@dlongley k, rebasing...

dmitrizagidulin:
Ok, tests pass locally, and should pass here now.

dmitrizagidulin:
@dlongley I still have the Karma/webpack issue, though, with `crypto` and `ursa` not being excluded.

webpack.config.js
```diff
@@ -94,7 +94,9 @@ outputs.forEach((info) => {
     externals: {
       'bitcore-message': 'bitcoreMessage',
       'node-forge': 'forge',
-      jsonld: 'jsonld'
+      jsonld: 'jsonld',
+      'crypto': false,
+      'ursa': false
```
dlongley: These don't need quotes, btw.

dlongley:
@dmitrizagidulin, do we have a build of crypto-ld that works in the browser already? If so, we should just do "externals" for `crypto-ld`, right?

dlongley:
@davidlehn, since you know more about the webpack magic, can you comment here? We need crypto-ld to work in the browser and we need to be able to run karma here using it, etc.

dmitrizagidulin:
@dlongley 
> do we have a build of crypto-ld that works in the browser already?

No, this is the first attempt to get it working in the browser.

dmitrizagidulin:
(Rebasing on `securityContext` PR).

package.json
```diff
@@ -36,6 +36,7 @@
     "bitcore-message": "github:CoMakery/bitcore-message#dist",
     "bs58": "^4.0.1",
     "chloride": "^2.2.8",
```
dlongley: Remove `chloride`

lib/jsonld-signatures.js
```diff
@@ -57,7 +57,7 @@ api.suites = require('./suites').suites;
 api.purposes = require('./purposes').purposes;
 
 // expose LDKeyPair classes
-Object.assign(api, require('./LDKeyPair'));
+// Object.assign(api, require('./LDKeyPair'));
```
dlongley: We need to expose the keypair stuff to avoid doing another major release.

```suggestion
  Object.assign(api, require('crypto-ld'));
```

package.json
```diff
@@ -36,6 +36,7 @@
...
+    "crypto-ld": "digitalbazaar/crypto-ld#signer",
```
dlongley: This also needs to use the new crypto-ld package, not a branch.

dmitrizagidulin:
@dlongley 
> can you quickly look into fixing the issues here?

Yep, on it!

package.json
dmitrizagidulin: Fixed.

package.json
dmitrizagidulin: Fixed.

PR #51, opened 2019-01-08
Fix proof matching.

mattcollier:
You'll let me know if the tests were not patched properly.

### DIFF
--- a/lib/ProofSet.js
+++ b/lib/ProofSet.js
@@ -306,9 +306,9 @@ async function _verify({
   document, suites, proofSet, purpose,
   documentLoader, expansionMap, compactProof}) {
   // filter out matching proofs
-  const matches = proofSet.filter(proof => purpose ?
-    purpose.match(proof, {document, documentLoader, expansionMap}) :
-    true);
+  const result = await Promise.all(proofSet.map(proof =>
+    purpose.match(proof, {document, documentLoader, expansionMap})));
+  const matches = proofSet.filter((value, index) => result[index]);
   if(matches.length === 0) {
     // no matches, nothing to verify
     return [];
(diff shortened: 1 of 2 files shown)


lib/ProofSet.js
```diff
@@ -306,9 +306,10 @@ async function _verify({
   document, suites, proofSet, purpose,
   documentLoader, expansionMap, compactProof}) {
   // filter out matching proofs
-  const matches = proofSet.filter(proof => purpose ?
+  const result = await Promise.all(proofSet.map(proof => purpose ?
```
dlongley: `purpose` is always defined so we can remove the conditional and just always call `purpose.match`. I think this was some noise left over from legacy code.

lib/ProofSet.js
```diff
@@ -306,9 +306,10 @@ async function _verify({
...
     purpose.match(proof, {document, documentLoader, expansionMap}) :
-    true);
+    true));
+  const matches = proofSet.filter((value, index) => result[index]);
```
dlongley: Why are we using `index`? Does the order change somehow?

mattcollier: we are filtering `proofSet` based on `result`.  Is there a better way?

lib/ProofSet.js
mattcollier: Done.

lib/ProofSet.js
dlongley: Hmm, I guess it's fine. I was thinking we would map to the `proof` or `null` based on the result of awaiting `match` during the `map()` call and then filter out based on `null` or not (like below). But I suppose this gets the job done fine.

PR #53, opened 2019-01-11
Security context

mattcollier:
@davidlehn I don't know what we're doing about the `fs` related stuff in the `security-context` module.

If something throws a `TypeError` for instance, it was not getting surfaced properly, this addresses that.

Also, on L258 `results.some` was being used.  It appears to me that every proof that is checked in `_verify` needs to be valid.  I believe the results from `_verify` are *only* the proofs matching the specified `purpose`.


lib/ProofSet.js
```diff
@@ -266,7 +267,7 @@ module.exports = class ProofSet {
       }
       return {verified, results};
     } catch(error) {
-      return {verified: false, error};
+      return {verified: false, error: serializeError(error)};
```
dlongley: I don't know why we're serializing errors like this -- it seems like an anti-pattern. It makes sense if we're, for example, returning the results via an HTTP API, but why lose native `Error` support?

lib/ProofSet.js
```diff
@@ -254,10 +255,10 @@ module.exports = class ProofSet {
       }
 
       // combine results
-      const verified = results.some(r => r.verified);
+      const verified = results.every(r => r.verified);
```
dlongley: I don't think this change is right, because:

The verify API tells you if "there is a proof that passes the required parameters" not "all proofs that could possibly match any of the parameters pass all of them". The latter seems practically impossible to fulfill and would make the API useless for many cases.

We're trying to encourage more targeted use of the API *for specific purposes* (hence all the proof purpose stuff). The old API was way too open-ended in this way and was essentially designed as a "verify all the proofs in any document using whatever stuff is installed in the system". That was decidedly the wrong decision -- so we've moved away to call verify to see if there's a proof on the document that matches the purpose of what the user is trying to do (and that matches the acceptable signature suite(s) and verification method(s) and so on).

mattcollier: I agree with what you said.  However `results` contains a list of proofs that *did* match the specified proofPurpose.  So, if I specify that the proofPurpose is `capabilityFoo`, I expect verify to validate *all* the `capabilityFoo` proofs right?  Unless I'm reading this wrong, with `results.some`it passes if only one (some) are valid.

lib/ProofSet.js
mattcollier: Maybe we don't want this behavior all the time, but having to relay these errors to HTTP is common enough that having to serialize the error with it's requisite deps in all the places is unattractive.  How do you feel about a `serializeError` boolean param to the verify method?

lib/ProofSet.js
dlongley: I'm not clear on what the use case is for requiring *all* proofs that match to verify. It seems that the more common case would be to pass if at least one passes ... I'm thinking of the case where perhaps a document was signed by three entities but one of them fails. Do we care if that proof is bogus? I would say we'd only care if those proofs are *linked* to one another (some kind of multiproof... but I suspect such a thing would be a different kind of proof that then wouldn't "match"), but otherwise, all we needed was one to pass.

lib/ProofSet.js
dlongley: Adding a special flag that will do it for you (and it's off by default) sounds good. Let's do that.

mattcollier: sounds good.

lib/ProofSet.js
mattcollier: The 2.0 API fails if any of the proofs fail.  This is pre-ocap.  Does ocap somehow alter the contract?

dlongley: @mattcollier, I'm saying the 2.0 API was broken and not useful for many cases. I don't think this has anything to do with ocap. This has to do with making the API more contextual -- which is both a good idea and a result of adding purposes and removing the whole "injector" pattern of installing all the possible things that one might possibly verify with.

mattcollier: OK, 2.0 API aside.  You gave the OR example for signatures A, B, or C.  What about the AND use case A, B, and C.  Seems like we must be explicit about whether it's an AND or an OR.  My expectation as a consumer would be that it defaults to an AND.

dlongley: @mattcollier, we explicitly moved all of those sorts of things outside of the API to make it simpler in 3.0. It was getting too unwieldy and awkward to use with all of the special flags and hooks. It's unclear how people might want to cobble various combinations together or how to use flags like "and" and "or". So we pushed all of that out to the edge -- let other people figure out what to do in their own way via processing the result or using custom proof purposes/suites.

> My expectation as a consumer would be that it defaults to an AND.

Why is that your expectation?

I actually think the opposite -- and am doing so based on our own use cases. They are all "or" cases to my knowledge so far. When using a proof "set", I would expect most people to be looking for "at least one" of the proofs to verify for their given parameters rather than needing all of them to. If we default to "AND" then I suspect we're going to be flipping that flag all the time to avoid throwing out perfectly good results.

If someone wants an "AND" they should check each individual result in the output or require a "multiproof" type of proof.

mattcollier: @msporny @davidlehn @dmitrizagidulin @gannan08 We are looking for additional opinions on this matter.

mattcollier:
Closing this, new PR will be based on latest `security-context` release.

PR #57, opened 2019-01-24
Update deps

davidlehn:
This updates to jsonld 1.5.0.  (It would be used anyway according to semver.)  Due to the jsonld changes, it will default to JS canonize now.  Is special support desired here to allow passing through the useNative flag?


dlongley:
> Is special support desired here to allow passing through the useNative flag?

Yes. We'd want that on the signature suites.

davidlehn:
@dlongley Is this the right useNative patch?  I think I got every spot?  Hard to test this, did some spot checks manually.  Seems kind of ugly to have to add this everywhere.

dlongley:
Let's do `useNativeCanonize`.

CHANGELOG.md
```diff
@@ -1,7 +1,11 @@
 # jsonld-signatures ChangeLog
 
 ### Changed
-- Updated jsonld to 1.5.0.
+- **NOTE**: Updated jsonld to 1.5.0. May require code updates to continue using
```
mattcollier: I would say `Code updates will be required if...`

And perhaps a link to specific documentation if available.

davidlehn:
@dmitrizagidulin You only use the native version if you need it for your use case, which your benchmarking will tell you.  It will depend on the use case, data size, data structure, js runtime, compiler, memory, parallelism, cores, etc etc.  I expect most people are fine with js with modern vms, but I don't know.  I'd rather not write a tome about it and just tell people to try and see.  Easy enough to do, and hard to otherwise know what to say.  PRs welcome to improve the text about this here, in jsonld.js, and/or in rdf-canonize.

PR #62, opened 2019-04-09
Make key signer/verifier factories optional.

mattcollier:
There is an allowance here that `this.verifier` may not be defined:

https://github.com/digitalbazaar/jsonld-signatures/blob/master/lib/suites/JwsLinkedDataSignature.js#L134

However, when using the constructor, it is expected that if a key is passed in as a parameter, that it *must* provide both a `signer` and `verifier` factory:

https://github.com/digitalbazaar/jsonld-signatures/blob/master/lib/suites/JwsLinkedDataSignature.js#L56

If the conditionals in this PR seem appropriate, it will probably be good to add some validation in the sign/verify APIs to ensure proper behavior when `this.signer` or `this.verifier` are undefined.

### DIFF
--- a/lib/suites/JwsLinkedDataSignature.js
+++ b/lib/suites/JwsLinkedDataSignature.js
@@ -52,8 +52,12 @@ module.exports = class JwsLinkedDataSignature extends LinkedDataSignature {
         }
       }
       this.key = key;
-      this.signer = key.signer();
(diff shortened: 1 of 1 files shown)


mattcollier:
@dlongley added one more commit here that needs approval.

Issue #61, opened 2019-04-09
Add more documentation about the usage of `documentLoader`.


gannan08:

dlongley:
To resolve this issue, we need to see updates to the README for the top-level `sign` and `verify` APIs that provide clear examples for how and when a `documentLoader` needs to be passed.

OR13:
I use this documentLoader in fixtures:

```
import jsonld from 'jsonld';
import contexts from './contexts';
import resolver from './resolver';

const nodeDocumentLoader = jsonld.documentLoaders.node();

export default (url: string, callback: any) => {
  // console.log(url);
  // are we handling a DID?
  if (url.indexOf('did:example') === 0) {
    const doc = resolver.resolve(url);
    if (!doc) {
      throw new Error('Could not resolve: ' + url);
    }
    // iterate public keys, find the correct id...
    for (const publicKey of doc.publicKey) {
      if (publicKey.id === url) {
        return callback(null, {
          contextUrl: null, // this is for a context via a link header
          document: publicKey, // this is the actual document that was loaded
          documentUrl: url, // this is the actual context URL after redirects
        });
      }
    }
  }

  //   are we handling a custom context?
  if (url in contexts) {
    const document = contexts[url];
    return callback(null, {
      contextUrl: null, // this is for a context via a link header
      document, // this is the actual document that was loaded
      documentUrl: url, // this is the actual context URL after redirects
    });
  }

  //   is this a published (public) context?
  return nodeDocumentLoader(url, callback);
};

```

With small amount of cleanup, it might help resolve this issue.

When I first started with JSON-LD I spent a lot of time being confused by this.

Since a lot of people are trying to work with DIDs, I think the example  / documentation should include some opinion on DID Resolvers.

OR13:
This is still an issue which is preventing wider adoption of linked data signatures in the DID community. You should provide an example.

OR13:
Be especially aware, that if you are relying on 

https://w3id.org/security/v1
https://w3id.org/security/v2

And using a custom document loader, you will need to make sure that these have been updated, to support any properties you are working with....

OR13:
also note that the example I provided above is not correct.... when loading a document of the form:

`did:example:123#WqzaOweASs78whhl_YvCEvj1nd89IycryVlmZMefcjU`

You should get the same result as:

`did:example:123`

... examples would really, really, really help here.

OR13:
something as simple as: 

```
[...]

PR #65, opened 2019-04-18
Add document loader docs

aljones15:
Adds documentation to documentLoader
Addresses:  https://github.com/digitalbazaar/jsonld-signatures/issues/61

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ This library's default `documentLoader` is very strict for security and content 
 - https://w3id.org/security/v1
 - https://w3id.org/security/v2
 
-If you require other security contexts you will need to provide a `documentLoader`
+If you require other documents to be loaded then you will need to provide a `documentLoader`
 that can fetch them. jsonld.js provides both a node and browser `documentLoader` you can use, however, depending on your use case, you may increase security by using a custom `documentLoader` that is similarly strict and will only load a subset of documents that is constrained by some technical, security, or business rules.
 
 Install with npm:
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ const publicKey = {
 const controller = {
   '@context': jsigs.SECURITY_CONTEXT_URL,
   id: 'https://example.com/i/alice',
-  publicKey: [publicKey]
+  publicKey: [publicKey],
   // this authorizes this key to be used for making assertions
   assertionMethod: [publicKey.id]
 };
@@ -108,7 +108,7 @@ const doc = {
 
 // sign the document as a simple assertion
 const {RsaSignature2018} = jsigs.suites;
-const {AuthenticationProofPurpose} = jsigs.purposes;
+const {AssertionProofPurpose} = jsigs.purposes;
 const {RSAKeyPair} = require('crypto-ld');
 const key = new RSAKeyPair({...publicKey, privateKeyPem});
 const signed = await jsigs.sign(doc, {
--- a/README.md
+++ b/README.md
@@ -110,6 +110,8 @@ const doc = {
 const {RsaSignature2018} = jsigs.suites;
 const {AssertionProofPurpose} = jsigs.purposes;
 const {RSAKeyPair} = require('crypto-ld');
+const {documentLoaders} = require('jsonld');
+
 const key = new RSAKeyPair({...publicKey, privateKeyPem});
 const signed = await jsigs.sign(doc, {
   suite: new RsaSignature2018({key}),
@@ -117,9 +119,12 @@ const signed = await jsigs.sign(doc, {
 });
 
 console.log('Signed document:', signed);
+// we will need the documentLoader to verify the controller
+const {node: documentLoader} = documentLoaders;
 
 // verify the signed document
 const result = await jsigs.verify(signed, {
+  documentLoader
   suite: new RsaSignature2018(key),
   purpose: new AssertionProofPurpose({controller})
 });
--- a/README.md
+++ b/README.md
@@ -39,13 +39,24 @@ This library allows for the key and key controller information to be looked up
 via a `documentLoader` or it can be provided directly to the API via the
 signature suite or proof purpose, respectively.
 
-This library's default `documentLoader` is very strict for security and content integrity purposes. It will only load locally available copies of the context documents that define the terms it uses internally. Any attempt to load any other documents (including other contexts) will throw an error. If other documents such as verification methods (e.g., public key documents), cannot be provided directly to the API and thus need to be loaded, a custom document loader must be passed. For the sake of clarity, the default document loader will only load locally available copies of the following documents:
+This library's default `documentLoader` is very strict for security and content
+integrity purposes. It will only load locally available copies of the context
+documents that define the terms it uses internally. Any attempt to load any
+other documents (including other contexts) will throw an error. If other
+documents such as verification methods (e.g., public key documents), cannot
+be provided directly to the API and thus need to be loaded, a custom document
+loader must be passed. For the sake of clarity, the default document loader
+will only load locally available copies of the following documents:
 
 - https://w3id.org/security/v1
 - https://w3id.org/security/v2
 
-If you require other documents to be loaded then you will need to provide a `documentLoader`
-that can fetch them. jsonld.js provides both a node and browser `documentLoader` you can use, however, depending on your use case, you may increase security by using a custom `documentLoader` that is similarly strict and will only load a subset of documents that is constrained by some technical, security, or business rules.
+If you require other documents to be loaded then you will need to provide a
+`documentLoader` that can fetch them. jsonld.js provides both a node and browser
+`documentLoader` you can use, however, depending on your use case, you may
+increase security by using a custom `documentLoader` that is similarly strict
+and will only load a subset of documents that is constrained by some technical,
+security, or business rules.
 
 Install with npm:
 
@@ -69,7 +80,9 @@ Signing and verifying a simple assertion:
 ```js
 // to generate the next two lines, run the following command:
 //
-// openssl genrsa -out key.pem; cat key.pem; openssl rsa -in key.pem -pubout -out pubkey.pem; cat pubkey.pem; rm key.pem pubkey.pem
+// openssl genrsa -out key.pem; cat key.pem; 
+// openssl rsa -in key.pem -pubout -out pubkey.pem;
+// cat pubkey.pem; rm key.pem pubkey.pem
 //
 // for an example of how to specify these keys, look at [key-example]:
 const publicKeyPem = "-----BEGIN PUBLIC KEY-----\r\n...";
@@ -124,7 +137,7 @@ const {node: documentLoader} = documentLoaders;
 
 // verify the signed document
 const result = await jsigs.verify(signed, {
-  documentLoader
+  documentLoader,
   suite: new RsaSignature2018(key),
   purpose: new AssertionProofPurpose({controller})
 });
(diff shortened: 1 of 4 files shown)


dlongley:
Just to be clear, this doesn't address #61 -- which wasn't clear at all in that issue. Issue #61 is about updating the README for the top-level `sign` and `verify` API and providing clear examples for how and when a `documentLoader` needs to be passed. That issue is not about the low-level utility API for `documentLoader.js`.

lib/documentLoader.js
```diff
@@ -6,10 +6,29 @@
 // load locally embedded contexts
 const contexts = require('./contexts');
 
+/**
+ * documentLoader is a set of methods used to fetch jsonld documents.
+ * Most of these methods are there to ensure jsonld contexts are loaded
+ * correctly.
```
dlongley:
```suggestion
 * from an in-memory, immutable copy of the context document to ensure content integrity.
```

lib/documentLoader.js
```diff
@@ -6,10 +6,29 @@
...
+ * @module documentLoader
+ */
 const api = {};
 module.exports = api;
 
 api.extendContextLoader = documentLoader => {
+  /**
+   * extendContextLoader takes in a documentLoader. It then
+   * returns a function that accepts a url. If the url is in the context
+   * cache, it returns the security context
```
dlongley:
```suggestion
   * map, it returns a copy of an in-memory, immutable context document
```

lib/documentLoader.js
```diff
@@ -24,5 +43,18 @@ api.extendContextLoader = documentLoader => {
 };
 
 api.strictDocumentLoader = api.extendContextLoader(url => {
+  /**
+   * strictDocumentLoader if the url for a context is in the cache
```
dlongley:
```suggestion
   * strictDocumentLoader if the url for a context is in the context map
```

lib/documentLoader.js
```diff
@@ -24,5 +43,18 @@ api.extendContextLoader = documentLoader => {
...
+   * this returns the security context otherwise
```
dlongley:
```suggestion
   * this returns a copy of an in-memory, immutable context document
```

lib/documentLoader.js
```diff
@@ -24,5 +43,18 @@ api.extendContextLoader = documentLoader => {
...
+   * it throws an error. This ensures no network
+   * calls are made and all contexts are those available locally.
```
dlongley:
```suggestion
   * calls are made so the only documents that can be loaded are locally available contexts.
```

lib/documentLoader.js
```diff
@@ -24,5 +43,18 @@ api.extendContextLoader = documentLoader => {
...
+   * @see documentLoader.extendContextLoader
+   *
+   * @param {string} url - A valid url to a jsonld context.
+   *
+   * @throws {Error} Always throws an error if the
+   * url is not in the context cache.
```
dlongley:
```suggestion
   * url is not in the context map (i.e., not a URL for a locally available context document).
```

lib/documentLoader.js
```diff
@@ -24,5 +43,18 @@ api.extendContextLoader = documentLoader => {
...
+   * @return {Object} A security context.
```
dlongley:
```suggestion
   * @return {Object} A JSON-LD RemoteDocument that is a copy of a locally available context.
```

README.md
```diff
@@ -39,6 +39,14 @@ This library allows for the key and key controller information to be looked up
 via a `documentLoader` or it can be provided directly to the API via the
 signature suite or proof purpose, respectively.
 
+This library's `documentLoader` already contains the security contexts for:
```
dlongley:
```suggestion
This library's default `documentLoader` is very strict for security and content integrity purposes. It will only load locally available copies of the context documents that define the terms it uses internally. Any attempt to load any other documents (including other contexts) will throw an error. If other documents such as verification methods (e.g., public key documents), cannot be provided directly to the API and thus need to be loaded, a custom document loader must be passed. For the sake of clarity, the default document loader will only load locally available copies of the following documents:
```

README.md
```diff
@@ -39,6 +39,14 @@ This library allows for the key and key controller information to be looked up
...
+
+- https://w3id.org/security/v1
+- https://w3id.org/security/v2
+
+If you require other security contexts you will need to provide a `documentLoader`
```
dlongley:
```suggestion
If you require other documents to be loaded then you will need to provide a `documentLoader`
```

README.md
```diff
@@ -39,6 +39,14 @@ This library allows for the key and key controller information to be looked up
...
+that can fetch them. jsonld.js provides both a node and browser `documentLoader` you can use.
```
dlongley:
```suggestion
that can fetch them. jsonld.js provides both a node and browser `documentLoader` you can use, however, depending on your use case, you may increase security by using a custom `documentLoader` that is similarly strict and will only load a subset of documents that is constrained by some technical, security, or business rules.
```

README.md
```diff
@@ -100,18 +108,23 @@ const doc = {
...
   purpose: new AssertionProofPurpose()
 });
 
 console.log('Signed document:', signed);
+// we will need the documentLoader to verify the controller
+const {node: documentLoader} = documentLoaders;
 
 // verify the signed document
 const result = await jsigs.verify(signed, {
+  documentLoader
```
mattcollier: missing comma.

lib/documentLoader.js
```diff
@@ -2,14 +2,38 @@
  * Copyright (c) 2018 Digital Bazaar, Inc. All rights reserved.
  */
 'use strict';
+/*eslint max-len: ["error", { "ignoreComments": true }]*/
 
 // load locally embedded contexts
 const contexts = require('./contexts');
 
+/**
+ * documentLoader is a set of methods used to dereference jsonld documents.
```
dlongley: Hmm, this is confusing. Here `documentLoader` refers to this little utility module, which is not a set of methods used to dereference jsonld documents. Rather, it is "A utility module that provides a set of functions for using or extending jsonld-signature's built-in JSON-LD document loader." We can add "To learn more about document loaders, see (put link to jsonld.js here) or [The JSON-LD API spec](https://www.w3.org/TR/json-ld11-api/#loaddocumentcallback)".

lib/documentLoader.js
```diff
@@ -2,14 +2,38 @@
...
+ * Documents stored at urls are usually dereferenced via https requests
+ * while DIDs can be dereferenced by a variety of means.
```
dlongley: We should strike this sentence -- we don't mention what "DIDs" are and it doesn't make any sense why it's here to the reader. If the reader needs to know more about document loaders, we should add a reference to jsonld.js or the JSON-LD API spec as above -- otherwise they are just going to think that `documentLoader` is this little utility file that happens to be named after something from an entirely different library (jsonld.js).

lib/documentLoader.js
```diff
@@ -2,14 +2,38 @@
...
+ * Most of these methods are there to ensure jsonld contexts are loaded
+ * from an in-memory, immutable copy of the context
+ * document to ensure content integrity.
```
dlongley: Can strike the above, as well.

lib/documentLoader.js
```diff
@@ -2,14 +2,38 @@
...
+ * @module documentLoader
+ */
 const api = {};
 module.exports = api;
 
 api.extendContextLoader = documentLoader => {
+  /**
+   * extendContextLoader takes in a documentLoader. It then
+   * returns a function that accepts a url. If the url is in the context
```
dlongley: I think this still needs a little bikeshedding, perhaps:

"extendContextLoader extends another JSON-LD document loader. Given a document loader to extend, this method will return a new document loader that will first check for a URL in jsonld-signature's built-in context map and, if not found, it will fall back to using the passed document loader. This utility method can be used to ensure that any local, in-memory, immutable context documents provided by jsonld-signatures will be used prior to using another document loader to load other documents."

Note that there's no guarantee that any "remote source" will ever be used here, so we should strike that.

lib/documentLoader.js
```diff
@@ -24,5 +48,21 @@ api.extendContextLoader = documentLoader => {
 };
 
 api.strictDocumentLoader = api.extendContextLoader(url => {
+  /**
+   * strictDocumentLoader if the url for a context is in the context map
+   * this returns a copy of an in-memory, immutable context document
```
dlongley: We should just define this in terms of `extendContextLoader` -- so we have the definition in one place. We can continue to say that the reason you'd use this method is if you don't want to load anything other than the built-in contexts provided by jsonld-signatures.

README.md
```diff
@@ -39,6 +39,14 @@ This library allows for the key and key controller information to be looked up
 via a `documentLoader` or it can be provided directly to the API via the
 signature suite or proof purpose, respectively.
 
+This library's default `documentLoader` is very strict for security and content integrity purposes. It will only load locally available copies of the context documents that define the terms it uses internally. Any attempt to load any other documents (including other contexts) will throw an error. If other documents such as verification methods (e.g., public key documents), cannot be provided directly to the API and thus need to be loaded, a custom document loader must be passed. For the sake of clarity, the default document loader will only load locally available copies of the following documents:
```
dlongley: This looks like a single line -- can we break it up accordingly? (80 char line limit) ... same for anything else here.

lib/documentLoader.js
aljones15: this kind of highlights a point I was making the use of documentLoader in this project creates confusion on which documentLoader is being used. Anyways, that's a better description so working on it.

dlongley: Well, users don't ever use this file. The utilities in this file are exported on the main API. This file is an internal implementation detail for organizational purposes. Perhaps even having a documentation `module` tag here is a bad idea.

README.md
```diff
@@ -39,6 +39,25 @@ This library allows for the key and key controller information to be looked up
...
+documents such as verification methods (e.g., public key documents), cannot
+be provided directly to the API and thus need to be loaded, a custom document
+loader must be passed. For the sake of clarity, the default document loader
+will only load locally available copies of the following documents:
+
+- https://w3id.org/security/v1
+- https://w3id.org/security/v2
+
+If you require other documents to be loaded then you will need to provide a
+`documentLoader` that can fetch them. jsonld.js provides both a node and browser
```
dlongley:
```suggestion
`documentLoader` that can provide them. jsonld.js provides both a node and browser
```

lib/documentLoader.js
```diff
@@ -3,13 +3,39 @@
...
 
+/*eslint max-len: ["error", { "ignoreComments": true }]*/
+
 // load locally embedded contexts
 const contexts = require('./contexts');
 
+/**
+ * A utility module that provides a set of functions for using or
+ * extending jsonld-signature's built-in JSON-LD document loader.
+ * @module documentLoader
```
dlongley: Can we just remove this module tag? This is an internal file, we don't need to expose docs on the module itself as if it's a publicly available thing. Only the methods matter.

```suggestion
```

lib/documentLoader.js
```diff
@@ -3,13 +3,39 @@
  */
 'use strict';
 
+/*eslint max-len: ["error", { "ignoreComments": true }]*/
+
 // load locally embedded contexts
 const contexts = require('./contexts');
 
+/**
+ * A utility module that provides a set of functions for using or
```
dlongley:
```suggestion
 * This is a utility module that provides a set of functions for using or
```

dlongley:
Couple more nits here and then we can pull this in!

PR #67, opened 2019-05-15
Remove ECDSA Signature algorithm.

aljones15:
- Remove ECDSA-2016 from suites.
- Remove all tests for it.
- Remove all mock data for ECDSA signatures.


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # jsonld-signatures ChangeLog
 
+## 4.2.0 -
+### Changed
```
dlongley: Needs empty line above `#Changed` and it should say `#Removed`. This is also a breaking change, so it will have to be version 5, unfortunately.

aljones15: no problem

mattcollier:
@davidlehn can you do travis magic to run CI on this again now that node-12 issues should be resolved?

I'd like to see this completed, apparently there are a number of security vulnerability in subdependencies of the bitcore stuff being ripped out here.

davidlehn:
I piled on more updates as long as I was reviewing.
Should we add a pointer to https://github.com/digitalbazaar/ecdsa-koblitz-signature-2016?

aljones15:
@dlongley @davidlehn @mattcollier is this good to go? Did I do something right?

mattcollier:
@davidlehn Is this ready to roll?

davidlehn:
@mattcollier I think we can release this.  It'll be a major version so 4.x users shouldn't have problems. I hadn't done it yet since I wanted to finish up the `ecdsa-koblitz-signature-2016` package first.  Was thinking it would be good to make sure we have the right jsigs APIs to use external suites like that.  And have a useful build process so sites like json-ld.org playground can easily load suites with script tags.  But we can address all this later with new jsigs version if needed.

dlongley:
Moving ahead with removing this. We'll deal with any *potential* issues with new versions as needed.

PR #68, opened 2019-05-16
Use security-context@3.

mattcollier:
node and karma tests are passing locally.

@davidlehn are there any webpack config changes required here?


davidlehn:
@mattcollier re webpack, it looks like the output is pulling in the security-context dist/module.js file correctly, so I don't think any changes are needed.

davidlehn:
I just noticed the browserify tests failed.  That is unfortunate.  Might be tricky to fix.

davidlehn:
Easiest browserify fix was to use esmify.  Unsure if we need to document that trick or not.

mattcollier:
Just diff'd the context files against what is on master in this repo and they are identical.  Proceeding with merge.

PR #75, opened 2019-08-10
Use security-context@4.

mattcollier:
watching: https://github.com/AnthonyRonning/vc-testing/issues/3


mattcollier:
@dlongley @davidlehn @dmitrizagidulin We're going to release this as a minor, is that correct?

dlongley:
> We're going to release this as a minor, is that correct?

+1

PR #90, opened 2020-02-07
Include purposeResult in verifyResult.

mattcollier:
30 some-odd tests in the test suite will need to be updated because they do a deep equal on the `verify` result which will now need to include `purposeResult`.

It is my hope that someone else has cycles to fix the test suite after the feature is finalized.

### DIFF
--- a/tests/test-common.js
+++ b/tests/test-common.js
@@ -4,7 +4,7 @@
 /* eslint-disable indent */
 module.exports = async function(options) {
 
-  'use strict';
+'use strict';
 
 const {assert, constants, jsigs, mock, suites, util} = options;
 const {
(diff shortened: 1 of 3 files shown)


mattcollier:
@dmitrizagidulin thanks for picking this up!  Do you have an estimate for completion time?  I'm having to use dev branches in several places to specify this patch, the sooner we can get this merged the better.

tests/test-common.js
```diff
@@ -4,949 +4,787 @@
...
+          const property = suite.legacy ? 'signature' : 'proof';
+          const expectedPurposeResult = {
+            Ed25519Signature2018: {
+              valid: true
+            },
+            RsaSignature2018: {
+              valid: true
+            },
+            EcdsaKoblitzSignature2016: {
+              controller: {
```
dmitrizagidulin: This `proofResult` section for the legacy suites doesn't quite look right. As in, why is the key type RsaVerificationKey2018, on line 384, etc.

tests/test-common.js
```diff
@@ -4,7 +4,7 @@
 /* eslint-disable indent */
 module.exports = async function(options) {
 
-'use strict';
+  'use strict';
```
dlongley: Why this change? It seems that there are a bunch of indentation changes in this PR that are not relevant. Are those mistakes? They seem to ignore the `eslint-disable indent` warning above. Can we back those out if they were not intentional?

davidlehn: Ditto.  Hard to review with the indent changes.  In the future make that it's own commit. 
 Alternatively, if we want to have proper indents here, then remove the above eslint pragma.

dmitrizagidulin: Review-wise it's easier, you can just use the https://github.com/digitalbazaar/jsonld-signatures/pull/90/files?w=1 / `?w=1` option. But yeah, let me see if I can reformat.

dmitrizagidulin: Out of curiosity, why are the tests wrapped in a `module.exports`?

aljones15: @dmitrizagidulin there is an options parameter so multiple option can be passed in. I actually think it's good design as we could start working on a json manifest file that setups test cases like the `json-ld` ones only in this case for different suites and keys etc.

dmitrizagidulin: @aljones15 But that's not what it's doing. It's just invoking the whole thing, once. (And iterating between suites /within/ that.)

aljones15: > @aljones15 But that's not what it's doing. It's just invoking the whole thing, once. (And iterating between suites /within/ that.)

I didn't write it, but yeah it is an enigma in our testing conventions. Regardless it does allow you to add new suites with out necessarily writing new tests etc. Perhaps the original author can defend it better.

davidlehn: The test wrapping was historically done in jsonld.js due to needing to pass in different apis for node vs browser testing.  It got copied here.  The main entry point test.js and test-karma.js files could pass in what was needed for the platform.  There may be better modern approaches if anyone wants to research that.

aljones15: > The test wrapping was historically done in jsonld.js due to needing to pass in different apis for node vs browser testing. It got copied here. The main entry point test.js and test-karma.js files could pass in what was needed for the platform. There may be better modern approaches if anyone wants to research that.

we actually might considering using it as it generates semantically meanful tests from json which is nice to have. I just wrote 2 test suites in the last 2 weeks and both iterate through 2 different `Suites` so it would have been better to just have a set test suite for each suite, pass in the necessary parameters and go.

tests/test-common.js
```diff
@@ -994,65 +1290,71 @@ describe('JSON-LD Signatures', () => {
...
+                },
+                id: 'https://example.com/i/alice',
+                publicKey: 'https://example.com/i/alice/keys/1'
+              },
+              valid: true
+            }
           };
-          assert.deepEqual(result, expected);
-        });
-        it('should sign and verify without a controller passed to purpose',
```
dmitrizagidulin: I don't know why this test disappeared, during rebasing. Investigating.

dmitrizagidulin:
@dlongley @davidlehn Ok, formatting reverted, should be good to go.

Issue #89, opened 2020-02-07
controllerId is overriden in ControllerProofPurpose


gjgd:
The problem I'm facing happens at this line: https://github.com/digitalbazaar/jsonld-signatures/blob/master/lib/purposes/ControllerProofPurpose.js#L59

If the controller is an object here, `controllerId` gets set to `controller.id` but then is overriden by the whole `controller` object.

This causes this line to fail later on: https://github.com/digitalbazaar/jsonld-signatures/blob/master/lib/purposes/ControllerProofPurpose.js#L72

I think the condition should look something like this instead:
```js
if(typeof controller === 'object') {
  controllerId = controller.id;
} else if(typeof controller !== 'string') {
  throw new TypeError("controller" must be a string representing a URL.');
} else {
  controllerId = controller;
}
```

aljones15:
@davidlehn @dlongley just to make sure controller can not be Array<string> of ids?
According to the [DiD Spec:](https://w3c.github.io/did-core/#dfn-did-controllers)

> DID controller
> The entity that has the ability to make changes to a DID document. A DID can have more than one controller. The DID controller(s) are denoted by the controller property on the top level of the DID document. Note that the DID controller(s) might include the DID subject.

So I believe in the case where a DiD has multiple controllers it might be an Array<string> or Array<object> Anyways @gjgd thanks for the issue and I am adding a unit test for this case right now. I will also add a commit with your fix, but it looks like it might need review as I think we might need to handle the case where a controller is an array, but I could be wrong.

aljones15:
@gjgd can you confirm if this is the error you are getting:

```
{ valid: false,
  error:
   { JsonLdError [jsonld.SyntaxError]: Invalid JSON-LD syntax; "@id" value an empty object or array of strings, if framing
       at _expandObject (/work/ajones/digital_bazaar/jsonld-signatures/node_modules/jsonld/lib/expand.js:480:19)
       at api.expand (/work/ajones/digital_bazaar/jsonld-signatures/node_modules/jsonld/lib/expand.js:246:9)
     name: 'jsonld.SyntaxError',
     details: { code: 'invalid @id value', value: [Object] } } }
```

gjgd:
@aljones15  That's correct

dlongley:
[...]

PR #91, opened 2020-02-07
Issue 89 overwritten controller

aljones15:
This corrects https://github.com/digitalbazaar/jsonld-signatures/issues/89

using the fixed suggested by @gjgd

Adds 4 new tests and 2 new keys.

```
--------------------------------|----------|----------|----------|----------|-------------------|
File                            |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
--------------------------------|----------|----------|----------|----------|-------------------|
All files                       |     79.9 |    65.77 |    85.57 |    79.64 |                   |
 lib                            |    74.09 |    62.31 |    78.72 |    73.41 |                   |
  ProofSet.js                   |    91.74 |    76.47 |    94.74 |     91.3 |... 29,238,330,335 |
  constants.js                  |      100 |      100 |      100 |      100 |                   |
  contexts.js                   |      100 |      100 |      100 |      100 |                   |
  documentLoader.js             |    90.91 |      100 |    66.67 |    90.91 |                66 |
  env.js                        |      100 |    66.67 |      100 |      100 |                 9 |
  expansionMap.js               |    66.67 |       50 |      100 |    66.67 |                 9 |
  jsonld-signatures.js          |    64.29 |    58.33 |      100 |    64.29 |... 41,42,43,46,47 |
  purposes.js                   |      100 |      100 |      100 |      100 |                   |
  suites.js                     |      100 |      100 |      100 |      100 |                   |
  util.js                       |    49.48 |     37.5 |    63.64 |    48.96 |... 23,225,227,228 |
 lib/purposes                   |    88.89 |    72.86 |      100 |    88.89 |                   |
  AssertionProofPurpose.js      |      100 |      100 |      100 |      100 |                   |
  AuthenticationProofPurpose.js |    90.48 |    76.47 |      100 |    90.48 |             15,18 |
  ControllerProofPurpose.js     |    86.84 |    64.29 |      100 |    86.84 |    29,54,56,63,65 |
  ProofPurpose.js               |    86.96 |       75 |      100 |    86.96 |          17,21,27 |
  PublicKeyProofPurpose.js      |      100 |      100 |      100 |      100 |                   |
 lib/suites                     |    82.95 |    65.44 |    88.89 |    82.88 |                   |
  EcdsaKoblitzSignature2016.js  |     91.3 |    57.14 |      100 |     91.3 |             23,45 |
  Ed25519Signature2018.js       |      100 |      100 |      100 |      100 |                   |
  GraphSignature2012.js         |     87.5 |    66.67 |      100 |     87.5 |             48,53 |
  JwsLinkedDataSignature.js     |    83.33 |    70.45 |      100 |    83.33 |... 26,133,169,182 |
  LinkedDataProof.js            |    57.14 |    33.33 |       50 |    57.14 |           9,25,38 |
  LinkedDataSignature.js        |    84.93 |    61.54 |       80 |    84.93 |... 58,263,279,293 |
  LinkedDataSignature2015.js    |    76.81 |    68.75 |      100 |    76.47 |... 65,166,167,171 |
  RsaSignature2018.js           |      100 |        0 |      100 |      100 |                32 |
--------------------------------|----------|----------|----------|----------|-------------------|
```

Expands test coverage of purposes by 3%.

### DIFF
--- a/lib/purposes/ControllerProofPurpose.js
+++ b/lib/purposes/ControllerProofPurpose.js
@@ -55,8 +55,9 @@ module.exports = class ControllerProofPurpose extends ProofPurpose {
           } else if(typeof controller !== 'string') {
             throw new TypeError(
               '"controller" must be a string representing a URL.');
+          } else {
+            controllerId = controller;
           }
-          controllerId = controller;
         } else if(owner) {
           if(typeof owner === 'object') {
             controllerId = owner.id;
@@ -66,7 +67,6 @@ module.exports = class ControllerProofPurpose extends ProofPurpose {
           }
           controllerId = owner;
         }
-
         // Note: `expansionMap` is intentionally not passed; we can safely drop
         // properties here and must allow for it
         const {'@graph': [framed = {}]} = await jsonld.frame(controllerId, {
--- a/lib/purposes/ControllerProofPurpose.js
+++ b/lib/purposes/ControllerProofPurpose.js
@@ -46,7 +46,6 @@ module.exports = class ControllerProofPurpose extends ProofPurpose {
       if(this.controller) {
         result.controller = this.controller;
       } else {
-        // support legacy `owner` property
         const {controller, owner} = verificationMethod;
         let controllerId;
         if(controller) {
@@ -58,6 +57,7 @@ module.exports = class ControllerProofPurpose extends ProofPurpose {
           } else {
             controllerId = controller;
           }
+        // support legacy `owner` property
         } else if(owner) {
           if(typeof owner === 'object') {
             controllerId = owner.id;
--- a/lib/purposes/ControllerProofPurpose.js
+++ b/lib/purposes/ControllerProofPurpose.js
@@ -46,6 +46,7 @@ module.exports = class ControllerProofPurpose extends ProofPurpose {
       if(this.controller) {
         result.controller = this.controller;
       } else {
+        // support legacy `owner` property
         const {controller, owner} = verificationMethod;
         let controllerId;
         if(controller) {
@@ -57,7 +58,6 @@ module.exports = class ControllerProofPurpose extends ProofPurpose {
           } else {
             controllerId = controller;
           }
-        // support legacy `owner` property
         } else if(owner) {
           if(typeof owner === 'object') {
             controllerId = owner.id;
@@ -80,7 +80,6 @@ module.exports = class ControllerProofPurpose extends ProofPurpose {
         }, {documentLoader, compactToRelative: false});
         result.controller = framed;
       }
-
       const verificationMethods = jsonld.getValues(
         result.controller, this.term);
       result.valid = verificationMethods.some(vm =>
--- a/lib/purposes/ControllerProofPurpose.js
+++ b/lib/purposes/ControllerProofPurpose.js
@@ -73,6 +73,8 @@ module.exports = class ControllerProofPurpose extends ProofPurpose {
         const {'@graph': [framed = {}]} = await jsonld.frame(controllerId, {
           '@context': constants.SECURITY_CONTEXT_URL,
           id: controllerId,
+          // the term should be in the json-ld object the controllerId resolves
+          // to.
           [this.term]: {
             '@embed': '@never',
             id: verificationId
--- a/lib/purposes/ControllerProofPurpose.js
+++ b/lib/purposes/ControllerProofPurpose.js
@@ -64,8 +64,9 @@ module.exports = class ControllerProofPurpose extends ProofPurpose {
           } else if(typeof owner !== 'string') {
             throw new TypeError(
               '"owner" must be a string representing a URL.');
+          } else {
+            controllerId = owner;
           }
-          controllerId = owner;
         }
         // Note: `expansionMap` is intentionally not passed; we can safely drop
         // properties here and must allow for it
--- a/tests/mock/Ed25519Signature2018.js
+++ b/tests/mock/Ed25519Signature2018.js
@@ -78,8 +78,8 @@ mock.parameters.sign = {
   })
 };
 
-// this links back to a key with an assertionMethod on it
-mock.parameters.assertionMethod = {
+// this links back to a key with a controller object
+mock.parameters.controllerObject = {
   creator: publicKeys.ned.id,
   date: '2018-02-13T21:26:08Z',
   key: new Ed25519KeyPair({
--- a/tests/mock/Ed25519Signature2018.js
+++ b/tests/mock/Ed25519Signature2018.js
@@ -78,6 +78,16 @@ mock.parameters.sign = {
   })
 };
 
+// this links back to a key with an assertionMethod on it
+mock.parameters.assertionMethod = {
+  creator: publicKeys.ned.id,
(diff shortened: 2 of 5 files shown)


tests/test-common.js
```diff
@@ -137,6 +140,33 @@ describe('JSON-LD Signatures', () => {
     });
   });
 
+  context('PublicKeyProofPurpose', async () => {
```
dlongley: The `PublicKeyProofPurpose` class is for backwards compatibility/deprecated proofs. We should at least test this with `AssertionProofPurpose` because it's a commonly used proof purpose. We should be able to just move it to where other tests of that purpose are then.

The verification method should also have the security v2 context on it (can be copied from the proof below).

lib/purposes/ControllerProofPurpose.js
```diff
@@ -66,7 +67,6 @@ module.exports = class ControllerProofPurpose extends ProofPurpose {
           }
           controllerId = owner;
```
dlongley: We need a similar fix here with owner.

aljones15: Done: https://github.com/digitalbazaar/jsonld-signatures/pull/91/commits/81d55746c846c55d9b96041f8489f30d15c2b570

tests/test-common.js
aljones15: > We should be able to just move it to where other tests of that purpose are then.
> 
> The verification method should also have the security v2 context on it (can be copied from the proof below).

This is where I ran into the issue. The AssertionProofPurpose block is hardwired to a key with an owner. We actually don't test any keys in any of the suites with `controller` or with `controller` and an object. Perhaps I can expand the keys to test for those scenarios?

dlongley: @aljones15, if by "expand the keys" you mean add more and run more tests (vs. change existing mock data/tests), then +1 to that. So -1 to changing any existing stuff, +1 to being additive.

aljones15: @dlongley  ok so basically I can copy the existing `sign and verify` test for `AssertionProofPurpose`, remove the `controller` being passed into the purpose, change the signer's id to point at a key with a controller object, and hope for the best. I think.

aljones15: Ok this is now done: https://github.com/digitalbazaar/jsonld-signatures/pull/91/commits/23a27c2a76f29aa59ed51b36bcc4ef7a65d61a29

and also here:

https://github.com/digitalbazaar/jsonld-signatures/pull/91/commits/6e33e9218dfd400e65e59eb1240eb00be1a78ba5

one weird thing I did notice: despite having public keys with:
```
controller: {id: 'foo'}
```

when it reaches the verificationMethod it has been reduced to just `controller: foo` so I'm not sure which Purpose @gjgd was using.

tests/mock/Ed25519Signature2018.js
```diff
@@ -78,6 +78,16 @@ mock.parameters.sign = {
   })
 };
 
+// this links back to a key with an assertionMethod on it
```
dlongley: What does "a key with an assertionMethod on it" mean? This comment needs adjusting (and the same comment appears in RsaSignature2018.js). Should this say "a key with a controller that is an object"?

tests/test-common.js
```diff
@@ -1006,6 +1008,48 @@ describe('JSON-LD Signatures', () => {
           };
           assert.deepEqual(result, expected);
         });
+        it('should sign and verify with out a controller passed to purpose',
```
dlongley:
```suggestion
        it('should sign and verify without a controller passed to purpose',
```

tests/test-common.js
```diff
@@ -1135,9 +1179,46 @@ describe('JSON-LD Signatures', () => {
           };
           assert.deepEqual(result, expected);
         });
+
+        it('should sign and verify with out a controller passed to purpose',
```
dlongley:
```suggestion
        it('should sign and verify without a controller passed to purpose',
```

tests/test-common.js
dlongley: There are already tests that sign and verify without a controller passed to purpose. Should this say: 'should sign and verify using a key that has a controller object'? Or something to that effect?

tests/test-common.js
```diff
@@ -1006,6 +1008,48 @@ describe('JSON-LD Signatures', () => {
           };
           assert.deepEqual(result, expected);
         });
+        it('should sign and verify with out a controller passed to purpose',
+          async () => {
+          const Suite = suites[suiteName];
+          const signSuite = new Suite({
+            ...mock.suites[suiteName].parameters.assertionMethod,
```
dlongley: If these are supposed to be params for an assertionMethod proof ... why is the proof purpose an AuthenticationProofPurpose below? Should the params be renamed to something like `keyWithControllerObject`?

aljones15: I changed it `controllerObject` hope that is descriptive enough: https://github.com/digitalbazaar/jsonld-signatures/pull/91/commits/c7f463c7169addf6bcdb0dec6f35f64b7994a3ac

tests/mock/Ed25519Signature2018.js
aljones15: Changed to `key with a controller object` https://github.com/digitalbazaar/jsonld-signatures/commit/c7f463c7169addf6bcdb0dec6f35f64b7994a3ac

aljones15:
@dlongley WRT to the name of the test itself I think you might be seeing this a tad out of context:

```
      AuthenticationProofPurpose
        ✓ should detect an expired date
        ✓ should detect a non-matching challenge
        ✓ should detect a non-matching domain
        ✓ should fail to verify because the purpose is not authorized
        ✓ should sign and verify
        ✓ should sign and verify without a controller passed to purpose
      AssertionProofPurpose
        ✓ should detect an expired date
        ✓ should fail to verify because the purpose is not authorized
        ✓ should sign and verify
        ✓ should sign and verify without a controller passed to purpose
    Current suite tests: RsaSignature2018
      ✓ should fail to verify a document w/public key missing a type
      AuthenticationProofPurpose
        ✓ should detect an expired date
        ✓ should detect a non-matching challenge
        ✓ should detect a non-matching domain
        ✓ should fail to verify because the purpose is not authorized
        ✓ should sign and verify
        ✓ should sign and verify without a controller passed to purpose
      AssertionProofPurpose
        ✓ should detect an expired date
        ✓ should fail to verify because the purpose is not authorized
        ✓ should sign and verify
        ✓ should sign and verify without a controller passed to purpose
```

in both the Authentication and Assertion Proof tests a `controller` is always passed in the passing tests so what makes this notable in those suites is that no `controller` is passed and instead when the controller is dereferenced the term used for the validation is found on the controller object.

aljones15:
> `should sign and verify with a controller string passed to purpose`
> `should sign and verify with a controller object passed to purpose`
> `should sign and verify without a controller passed to purpose using a key with a controller string`
> `should sign and verify without a controller passed to purpose using a key with a controller object`

Can I add those to:
https://github.com/digitalbazaar/jsonld-signatures/issues/92

and we can go ahead and fix this bug?

PR #93, opened 2020-02-12
Wrap multiple verify() errors in an umbrella error.

dmitrizagidulin:
This PR ensures that verify's result.error always contains just one error.
In the case of multiple errors, they're wrapped in a top-level error (and the actual list is contained in error.cause).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@
   - Remove bitcore-message dependency.
   - Move feature to 'ecdsa-koblitz-signature-2016' package.
 
+### Changed
+  - **BREAKING**: `verify()`'s results.error is now always a 
+  `VerificationError` instance where `error.errors` is an array that includes 
+  all of the errors that occurred during the verification process.
+
 ### 4.6.0 - 2020-01-17
 
 ### Changed
--- a/lib/VerificationError.js
+++ b/lib/VerificationError.js
@@ -1,0 +1,20 @@
+/*!
+ * Copyright (c) 2020 Digital Bazaar, Inc. All rights reserved.
+ */
+'use strict'
+
+/**
+ * Used as an umbrella wrapper around multiple verification errors.
+ */
+class VerificationError extends Error {
+  /**
+   * @param {Error|Error[]} errors
+   */
+  constructor(errors) {
+    super('Verification error(s).');
+
+    this.name = 'VerificationError';
+    this.errors = [].concat(errors);
+  }
+}
+module.exports = VerificationError;
--- a/lib/jsonld-signatures.js
+++ b/lib/jsonld-signatures.js
@@ -13,6 +13,7 @@ Object.assign(api, constants);
 
 // TODO: support `ProofChain`
 const ProofSet = require('./ProofSet');
+const VerificationError = require('./VerificationError');
 
 api.sign = async function sign(document, {
   suite, purpose, documentLoader, expansionMap, compactProof} = {}) {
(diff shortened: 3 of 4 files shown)


davidlehn:
I think it would be better if the top level Error was just something simple like 'Verification error.', and the the error or errors are normalized into an array as `err.errors`.  Will likely need docs to say that's how the API works.  Also don't forget to update the changelog with these sorts of changes.

dmitrizagidulin:
@davidlehn Sure thing. Changed it to `.errors`, and added a Changelog entry.

lib/jsonld-signatures.js
```diff
@@ -36,15 +36,25 @@ api.verify = async function verify(document, {
...
       const err = new Error(
         `A URL "${url}" could not be fetched; you need to pass ` +
         '"documentLoader" or resolve the URL before calling "verify".');
       err.cause = error;
       result.error = err;
+    } else if(Array.isArray(error)) {
+      const firstError = error[0];
+
+      if(error.length > 1) {
+        const err = new Error('Verification error: ' + firstError);
```
gannan08: What are we trying to achieve here? It is my understanding that we are trying to handle the case where we have multiple errors. If that is true, then it would make sense to me that the encapsulating Error message was along the lines of "Multiple errors found during the verification process." The Error name would be, if used elsewhere, "VerificationError". Thoughts?

@dmitrizagidulin @davidlehn

aljones15: As it stands the current code will take the message and name from firstError and append it to `'Verification error: '
```js
 var e = new Error('foo')
e.name
// 'Error'
e.message
// 'foo'
'VerificationError: ' + e
'VerificationError: Error: foo'
```

so if the error is an array then:

```js
const err = new Error(`Verification error: ${error.map(e => '' + e).join(', ')}`);
err.message
// Error Verification error: Error: foo, Error: foo
```

Will result in all the error messages being concatenated into a comma separated string.

Not sure if that's what we want here, but it is possible to summarize all the errors so far.
The other options would be push the actual errors into the details of the error i.e. the pattern used in bedrock.

aljones15: @gannan08 @dmitrizagidulin the `http-signature-zcap-verify` library has a helper function for validation errrors that might be a good pattern to use:

https://github.com/digitalbazaar/http-signature-zcap-verify/blob/a9fb5f05d4abab2290157d71f2a7b4c9ff15be94/main.js#L161-L167

the original error is stuffed in the `cause`  property.

dmitrizagidulin: @aljones15 Thanks. I briefly considered appending all the error messages, but that's not necessary, the first one will suffice.

lib/jsonld-signatures.js
```diff
@@ -36,15 +36,25 @@ api.verify = async function verify(document, {
   suite, purpose, documentLoader, expansionMap, compactProof} = {}) {
   const result = await new ProofSet().verify(
     document, {suite, purpose, documentLoader, expansionMap, compactProof});
-  if(!documentLoader && result.error) {
-    const {error} = result;
-    if(error.name === 'jsonld.InvalidUrl') {
+  const {error} = result;
+  if(error) {
+    if(!documentLoader && error.name === 'jsonld.InvalidUrl') {
```
davidlehn: Looks like in the array-of-errors case this is checking .name of an array.  I imagine that would work, but it's strange.

davidlehn:
Latest version behavior:
- One "jsonld.InvalidUrl" error: gets wrapped with some text and real error set as err.cause.
- Error array of 1 errors: error returned
- Error array of >1 errors: wrapped with "multiple" text, has a VerificationError .name, and has errors as err.errors.
- Changelog/docs do not reflect this complex behavior.
- I'm not sure what the test coverage is for all this.

If this is just for random human usage, maybe it doesn't matter.  If code has to deal with this, it might be easier to have:
- Single normalized error pattern.
- Always have top level "Verification errors.".
- err.errors that is always an array and has 1 or more errors.

dmitrizagidulin:
@davidlehn Changed to what you proposed.

lib/VerificationError.js
```diff
@@ -0,0 +1,20 @@
...
+'use strict'
+
+/**
+ * Used as an umbrella wrapper around multiple verification errors.
+ */
+class VerificationError extends Error {
+  /**
+   * @param {Error|Error[]} errors
+   */
+  constructor (errors) {
```
dlongley:
```suggestion
  constructor(errors) {
```

CHANGELOG.md
```diff
@@ -8,6 +8,11 @@
   - Remove bitcore-message dependency.
   - Move feature to 'ecdsa-koblitz-signature-2016' package.
 
+### Changed
+  - **BREAKING**: `verify()`'s results.error is now always an error, not an
+    array of errors. In case of multiple errors, a wrapping error is thrown,
```
dlongley: My understanding is that a wrapping error is always thrown (which +1 to that consistency, btw). So this text should say something more like: "results.error is now always a `VerificationError` instance where `error.errors` is an array that includes all of the errors that occurred during the verification process."

dmitrizagidulin: Thanks, changed!

PR #114, opened 2020-09-29
Drop support for Node 8.

mattcollier:
CI is failing on this PR because evidently sodium-native (by way of crypto-ld) no longer builds on node 8.  https://github.com/digitalbazaar/jsonld-signatures/pull/113

I am aware there is work on this branch that updates to crypto-ld@4 which l believe addresses this issue as well by eliminating the sodium-native dependency: https://github.com/digitalbazaar/jsonld-signatures/tree/v6.x

It would be good to get a release out that includes the node_forge update referenced in the security alert.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,10 +14,10 @@ script:
 matrix:
   include:
   - name: "Browser Unit Tests (webpack)"
-    node_js: "12"
+    node_js: "14"
     env: BUNDLER=webpack
   - name: "Browser Unit Tests (browserify)"
-    node_js: "12"
+    node_js: "14"
     env: BUNDLER=browserify
 notifications:
   email:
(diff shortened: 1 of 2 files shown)


dmitrizagidulin (approved):
👍 
Thanks @mattcollier

dlongley (approved):

davidlehn:
So sodium-native stopped building, and that effectively breaks any dependencies?  I guess they didn't or couldn't keep old node 8 support working as is?  Is this going to be a breaking change for this package since we are dropping support for node 8?  ... PR merged while I was typing.  I was going to suggest adding a changelog to the PR.  Otherwise it's fine.  Also seems like it would be possible to re-add 8 support once sodium-native is gone?  Probably not worth the trouble for EOL nodes.

PR #122, opened 2021-01-22
Update jsonld.

davidlehn:


dlongley (approved):
Thanks!

dlongley (changes requested):
Actually, looks like tests aren't passing.

dmitrizagidulin:
This PR has also now been tested downstream in vc-js v6.4, and passes tests. Merging.

PR #132, opened 2021-03-19
Use the document's context for proof before security context.

dmitrizagidulin:
Relevant for the 2020 sig suite.


dlongley (approved):
Needs a changelog update and a patch release. Thanks!

PR #137, opened 2021-04-06
Fix signer param in LinkedDataSignature constructor.

### LINKED ISSUE #136: Ensure LinkedDataSignature constructor initializes this.signer
Subclasses (like the 2020 suite) sometimes pass a `signer` object to the suite constructor, instead of a key.
Make sure that gets initialized in the constructor (this is currently not working).

mattcollier:
Fixes #136.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,15 @@
 # jsonld-signatures ChangeLog
 
-## 8.0.3 - TBD
+## 9.0.0 - TBD
+
+### Changed
+- **BREAKING**: Remove `verificationMethod` param from suite constructor. It
+  is now strictly initialized from `key.id` or `signer.id`.
+  Increase validation on either key or signer/verifier parameters.
 
 ### Fixed
-- Add missing `signer` parameter to the `LinkedDataSignature` constructor.
-  This issue caused `this.signer` in subclasses to be `undefined`.
+- Add missing `signer` and `verifier` parameters to the `LinkedDataSignature` 
+  constructor. This issue caused `this.signer` in subclasses to be `undefined`.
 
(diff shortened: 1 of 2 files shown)


mattcollier:
@dmitrizagidulin now it's the case that if one passes in a `signer` without a `verificationMethod`, a proof with a `verificationMethod` of `undefined` is produced. Parameter validation is needed to prevent this from happening.

I'm wondering what the relationship between the `signer`, `verificationMethod` and `key` parameters should be.

Is it the case that if a `signer` is specified a `verificationMethod` must also be specified **OR** a `key` is specified.  Which is to say, one may not pass in all three?

https://github.com/digitalbazaar/jsonld-signatures/blob/3f7be0f82e6e4f7bf0c1b33209a6b257e2f75765/lib/suites/LinkedDataSignature.js#L31-L52

dlongley:
I thought it used to be the case that `signer` included an `id` property that would be used in the `verificationMethod` field. Has that changed?

dlongley (approved):

dmitrizagidulin:
@dlongley 
> I thought it used to be the case that signer included an id property that would be used in the verificationMethod field. Has that changed?

The closest thing to that would be  https://github.com/digitalbazaar/ed25519-signature-2020/blob/main/lib/Ed25519Signature2020.js#L44, which uses the `key`'s id, but not the signer's.
We can modify it to also use the signer's.

dmitrizagidulin (approved):
:+1:

dmitrizagidulin:
@mattcollier @dlongley - I've added this logic here: https://github.com/digitalbazaar/jsonld-signatures/pull/138

PR #138, opened 2021-04-06
Add additional validation for verificationMethod, signer and verifier.

dmitrizagidulin:
An addition to PR https://github.com/digitalbazaar/jsonld-signatures/pull/137.

### DIFF
--- a/lib/suites/LinkedDataSignature.js
+++ b/lib/suites/LinkedDataSignature.js
@@ -54,7 +54,7 @@ module.exports = class LinkedDataSignature extends LinkedDataProof {
     super({type});
     this.LDKeyClass = LDKeyClass;
     this.proof = proof;
-    const vm = this._validateVerificationMethod({
+    const vm = this._processSignatureParams({
       verificationMethod, key, signer, verifier
     });
     this.verificationMethod = vm.verificationMethod;
@@ -68,53 +68,6 @@ module.exports = class LinkedDataSignature extends LinkedDataProof {
       }
     }
     this.useNativeCanonize = useNativeCanonize;
-  }
-
-  /**
-   * See constructor docstring for param details.
-   *
-   * @returns {{verificationMethod: string, key: LDKeyPair, signer: Function,
-   *   verifier: Function}} - Validated and initialized key-related parameters.
-   */
-  _validateVerificationMethod({verificationMethod, key, signer, verifier}) {
-    if(!(key || signer || verifier)) {
-      // eslint-disable-next-line max-len
-      throw new TypeError('Either a "key" or at least one of "signer"/"verifier" parameters is required');
-    }
-    if((key && signer) || (key && verifier)) {
-      // eslint-disable-next-line max-len
-      throw new TypeError('Either a "key" or at least one of "signer"/"verifier" parameters is required, but not both.');
-    }
-
-    const vm = {};
-    if(key) {
-      vm.key = key;
-      vm.verificationMethod = verificationMethod || key.id;
-      if(typeof key.signer === 'function') {
-        vm.signer = key.signer;
-      }
-      if(typeof key.verifier === 'function') {
-        vm.verifier = key.verifier;
-      }
-      if(!(vm.signer || vm.verifier)) {
-        // eslint-disable-next-line max-len
-        throw new TypeError('The "key" parameter must contain a "signer" or "verifier" method.');
-      }
-    } else {
-      vm.verificationMethod = verificationMethod ||
-        signer && signer.id ||
-        verifier && verifier.id;
-      vm.signer = signer;
-      vm.verifier = verifier;
-    }
-
-    if(!(vm.verificationMethod && typeof vm.verificationMethod !== 'string')) {
-      // This means that verificationMethod was not passed in directly,
-      // AND that neither key.id or signer.id/verifier.id exists.
-      throw new TypeError('"verificationMethod" must be a URL string.');
-    }
-
-    return vm;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


lib/suites/LinkedDataSignature.js
```diff
@@ -50,6 +70,53 @@ module.exports = class LinkedDataSignature extends LinkedDataProof {
     this.useNativeCanonize = useNativeCanonize;
   }
 
+  /**
+   * See constructor docstring for param details.
+   *
+   * @returns {{verificationMethod: string, key: LDKeyPair, signer: Function,
+   *   verifier: Function}} - Validated and initialized key-related parameters.
+   */
+  _validateVerificationMethod({verificationMethod, key, signer, verifier}) {
```
dlongley, line 79: Since this is private, please move it to the bottom of the class. Possibly rename to something like `processSignatureParams` as well.

lib/suites/LinkedDataSignature.js
```diff
@@ -11,37 +11,57 @@ const LinkedDataProof = require('./LinkedDataProof');
...
-    }
     super({type});
-    this.verificationMethod = verificationMethod;
     this.LDKeyClass = LDKeyClass;
-    this.signer = signer;
     this.proof = proof;
-    if(date !== undefined) {
+    const vm = this._validateVerificationMethod({
+      verificationMethod, key, signer, verifier
+    });
```
mattcollier, line 59: seem weird to return `vm` from a `validate` API which normally reports `{valid: true}`.  Can this pattern be used here or does this helper need to be renamed??

lib/suites/LinkedDataSignature.js
```diff
@@ -50,6 +70,53 @@ module.exports = class LinkedDataSignature extends LinkedDataProof {
...
 
+  /**
+   * See constructor docstring for param details.
+   *
+   * @returns {{verificationMethod: string, key: LDKeyPair, signer: Function,
+   *   verifier: Function}} - Validated and initialized key-related parameters.
+   */
+  _validateVerificationMethod({verificationMethod, key, signer, verifier}) {
+    if(!(key || signer || verifier)) {
+      // eslint-disable-next-line max-len
```
mattcollier, line 81: -1 to disabling linting in this use case.

lib/suites/LinkedDataSignature.js
```diff
@@ -11,37 +11,57 @@ const LinkedDataProof = require('./LinkedDataProof');
...
-   *   the `proof` node (e.g. any other custom fields can be provided here
-   *   using a context different from security-v2).
-   * @typedef {Object} LDKeyPair
-   * @param {LDKeyPair} LDKeyClass - The crypto-ld key type that this suite
-   *   will use to sign/verify signatures.
-   * @param [date] {string|Date} signing date to use if not passed.
-   * @param {function} [signer] - Signer function.
-   * @param [useNativeCanonize] {boolean} true to use a native canonize
-   *   algorithm.
+   * @param {string} [options.verificationMethod]  A key id URL of the paired
```
dlongley, line 39: Can we say when this would be passed in? It would only be used on the verification side if you've retrieved the verification method previously and don't need it to be looked up using the `documentLoader`, right?

dlongley (commented):
This is helpful, thanks.

lib/suites/LinkedDataSignature.js
```diff
@@ -50,6 +70,53 @@ module.exports = class LinkedDataSignature extends LinkedDataProof {
...
+  /**
+   * See constructor docstring for param details.
+   *
+   * @returns {{verificationMethod: string, key: LDKeyPair, signer: Function,
+   *   verifier: Function}} - Validated and initialized key-related parameters.
+   */
+  _validateVerificationMethod({verificationMethod, key, signer, verifier}) {
+    if(!(key || signer || verifier)) {
+      // eslint-disable-next-line max-len
+      throw new TypeError('Either a "key" or at least one of "signer"/"verifier" parameters is required');
```
dlongley, line 82:
```suggestion
      throw new TypeError(
        'Either a "key" or at least one of "signer"/"verifier" parameters is required.');
```

dlongley, line 82:
```suggestion
      throw new TypeError('Either "key" or one of "signer"/"verifier" is required.');
```

lib/suites/LinkedDataSignature.js
dmitrizagidulin: Yeah, let's rename the helper. (I'll use Dave's suggestion below.)

lib/suites/LinkedDataSignature.js
dmitrizagidulin: K, will fix. (But I would like to have a discussion about it over at https://github.com/digitalbazaar/eslint-config-digitalbazaar/issues/50 )

lib/suites/LinkedDataSignature.js
dmitrizagidulin: @dlongley hmm so honestly, I can't think of when it would be useful, on the verification side. Because if you've already retrieved it, the easiest thing is to just pass it in as a `key` param.

lib/suites/LinkedDataSignature.js
dmitrizagidulin: @dlongley - I think the 'at least' part is important, since you can pass in both. No?

lib/suites/LinkedDataSignature.js
dmitrizagidulin: Done (moved, renamed)

lib/suites/LinkedDataSignature.js
dmitrizagidulin: Fixed.

lib/suites/LinkedDataSignature.js
dmitrizagidulin: Renamed.

dmitrizagidulin:
@dlongley @mattcollier - I removed `verificationMethod` as a standalone suite constructor param. (Now it is always derived from `key.id` or `signer.id`.

lib/suites/LinkedDataSignature.js
```diff
@@ -281,4 +301,51 @@ module.exports = class LinkedDataSignature extends LinkedDataProof {
...
+      vm.verificationMethod = key.id;
+      if(typeof key.signer === 'function') {
+        vm.signer = key.signer();
+      }
+      if(typeof key.verifier === 'function') {
+        vm.verifier = key.verifier();
+      }
+      if(!(vm.signer || vm.verifier)) {
+        throw new TypeError('The "key" parameter must contain a "signer" or ' +
+          '"verifier" method.');
```
dlongley, line 330:
```suggestion
        throw new TypeError(
          'The "key" parameter must contain a "signer" or "verifier" method.');
```

lib/suites/LinkedDataSignature.js
```diff
@@ -281,4 +301,51 @@ module.exports = class LinkedDataSignature extends LinkedDataProof {
...
+      }
+    } else {
+      vm.verificationMethod = signer && signer.id || verifier && verifier.id;
```
dlongley, line 333:
```suggestion
      vm.verificationMethod = (signer && signer.id) || (verifier && verifier.id);
```

dlongley (approved):
Presuming any other comments have been addressed now, LGTM.

lib/suites/LinkedDataSignature.js
dmitrizagidulin: Changed.

lib/suites/LinkedDataSignature.js
dmitrizagidulin: Changed, thanks.

PR #139, opened 2021-04-08
Add a 'addSuiteContext' flag to LinkedDataSignature suite.

dmitrizagidulin:
Adding the flag to support the feature we discussed. (The actual enforcement mechanism will be on the suite level.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,15 @@
 
 ## 9.0.1 -
 
+These changes were intended to be released in v9.0.0, so, releasing them as
+a patch.
+
 ### Changed
-- Implement automatic adding of the suite context to the document to be signed,
-  (if it's not present already). (We meant to add this to v9.0.0, so we're
-  releasing this as a patch).
+- **BREAKING**: Implement automatic adding of the suite context to the document
+  to be signed, (if it's not present already).
+- **BREAKING**: Remove the case where the `document` argument in `sign()` or
+  `verify()` is a URL (instead of an object), since this is an unused feature,
+  and a mixing of layers.
 
 ## 9.0.0 - 2021-04-06
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # jsonld-signatures ChangeLog
 
+## 9.0.1 -
+
+### Changed
+- Implement automatic adding of the suite context to the document to be signed,
+  (if it's not present already). (We meant to add this to v9.0.0, so we're
+  releasing this as a patch).
+
 ## 9.0.0 - 2021-04-06
 
 ### Changed
--- a/lib/jsonld-signatures.js
+++ b/lib/jsonld-signatures.js
@@ -19,9 +19,7 @@ const VerificationError = require('./VerificationError');
  * Cryptographically signs the provided document by adding a `proof` section,
  * based on the provided suite and proof purpose.
  *
- * @param {object|string} document - The document to be signed, either
- *   a string URL (resolved to an object via the given `documentLoader`) or a
- *   plain object (JSON-LD document).
+ * @param {object} document - The JSON-LD document to be signed.
  *
  * @param {object} options - Options hashmap.
  * @param {LinkedDataSignature} options.suite - The linked data signature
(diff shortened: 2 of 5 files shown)


dlongley (approved):

dmitrizagidulin:
(This is tested in PR https://github.com/digitalbazaar/ed25519-signature-2020/pull/6 )

lib/jsonld-signatures.js
```diff
@@ -121,3 +135,44 @@ api.purposes = require('./purposes').purposes;
...
+      // Enforce the suite's context by adding it to the document
+      document['@context'] = [...document['@context'] || [], contextUrl];
+    } else {
+      throw new TypeError(
+        `The document to be signed must contain this suite's @context, ` +
+        `"${contextUrl}".`
+      );
+    }
+  }
+  // document does contain the context, do nothing
```
dlongley, line 161: Preferred pattern for reducing nesting and getting error conditions/early returns out of the way first:

```suggestion
  if(_includesContext({document, contextUrl})) {
    // document already includes the required context
    return;
  }

  if(!addSuiteContext) {
    throw new TypeError(
      `The document to be signed must contain this suite's @context, ` +
      `"${contextUrl}".`);
  }

  // enforce the suite's context by adding it to the document
  document['@context'] = [...document['@context'] || [], contextUrl];
```

dlongley (approved):

lib/jsonld-signatures.js
```diff
@@ -19,31 +19,47 @@ const VerificationError = require('./VerificationError');
  * Cryptographically signs the provided document by adding a `proof` section,
  * based on the provided suite and proof purpose.
  *
- * @param {object|string} document - The document to be signed, either a string
- *  URL (resolved to an object via the given `documentLoader`) or a plain
- *  object (JSON-LD document).
+ * @param {object|string} document - The document to be signed, either
```
mattcollier, line 22: I see here that the `document` can be a document OR a URL.  This PR does not address the case where `document` is a string at all.  Not only will the context not be added properly in that case, but I think `_ensureSuiteContext` breaks if `document` is a string.

I have a suspicion that we have no tests that involve `document` being a string.

`ProofSet().add` on L48 uses the `documentLoader` to dereference the URL as needed.

https://github.com/digitalbazaar/jsonld-signatures/blob/57a9f5b473f501864848fb29737cb96a7ed97fdf/lib/ProofSet.js#L66-L70

So, it seems to me that `_ensureSuiteContext` needs to be moved into `ProofSet().add()` after the existing documentLoader logic or the document loader business needs to move forward in the order of events.

mattcollier (changes requested):

lib/jsonld-signatures.js
mattcollier: And looking deeper into `ProofSet().add()`, maybe `_ensureSuiteContext` belongs back in the suite?

in `suite.createProof`:

https://github.com/digitalbazaar/jsonld-signatures/blob/57a9f5b473f501864848fb29737cb96a7ed97fdf/lib/ProofSet.js#L80-L83

dlongley: I don't think we support the document being a string. I don't think we ever have. We should remove that.

lib/jsonld-signatures.js
```diff
@@ -19,31 +19,47 @@ const VerificationError = require('./VerificationError');
...
 api.sign = async function sign(document, {
-  suite, purpose, documentLoader, expansionMap
+  suite, purpose, documentLoader, expansionMap, addSuiteContext = true
 } = {}) {
+  // Ensure document contains the signature suite specific context url
+  // or throw an error (in case an advanced user overrides the `addSuiteContext`
+  // flag to false).
+  _ensureSuiteContext({
+    document, contextUrl: suite.contextUrl, addSuiteContext
+  });
```
dlongley, line 61: Yeah, this will need to be moved into `ProofSet.add` to handle the document as a string use case.

lib/jsonld-signatures.js
dmitrizagidulin: Thanks, fixing!

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # jsonld-signatures ChangeLog
 
+## 9.0.1 -
+
+### Changed
+- Implement automatic adding of the suite context to the document to be signed,
+  (if it's not present already). (We meant to add this to v9.0.0, so we're
+  releasing this as a patch).
```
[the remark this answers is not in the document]
dmitrizagidulin: Thanks, rephrased.

lib/jsonld-signatures.js
dmitrizagidulin: Fixed.

dmitrizagidulin:
After further consultation, we've decided to remove the '`document` can be a URL' feature from `sign()` and `verify()`.

mattcollier (approved):
TY.

PR #140, opened 2021-04-10
Add support for dealing with credentials/v1 and ed25519-2018 contexts.

dmitrizagidulin:


dlongley:
This code needs to be suite specific -- so it can be put in the corner where it belongs :). If we need to expose some API to allow for suites to do more custom/fancy things in jsigs, that's fine, but the specifics should all be in the legacy suites that have this problem.

dlongley (changes requested):
This code needs to be suite specific -- so it can be put in the corner where it belongs :). If we need to expose some API to allow for suites to do more custom/fancy things in jsigs, that's fine, but the specifics should all be in the legacy suites that have this problem.

dmitrizagidulin:
@dlongley - Fixed (moved the checking of suite-specific context to suite instance method instead of top-level jsigs.

lib/suites/LinkedDataSignature.js
```diff
@@ -278,7 +278,7 @@ module.exports = class LinkedDataSignature extends LinkedDataProof {
     }
 
     // ensure verification method has not been revoked
-    if(framed.revoked !== undefined) {
+    if(framed.revoked) {
```
dlongley, line 281: Why the change here? I'd prefer to leave it how it was.

dlongley (approved):
Thanks! One minor quibble ... let's keep that revoked flag how it was.

lib/suites/LinkedDataSignature.js
dmitrizagidulin: No problem! Reverting.

lib/suites/LinkedDataSignature.js
```diff
@@ -311,8 +311,51 @@ module.exports = class LinkedDataSignature extends LinkedDataProof {
...
+        `"${contextUrl}".`);
+    }
+
+    // enforce the suite's context by adding it to the document
+    document['@context'] = [...document['@context'] || [], contextUrl];
+  }
 };
 
+/**
+ * Tests whether a provided JSON-LD document includes a context url in its
```
mattcollier, line 344:
```suggestion
 * Tests whether a provided JSON-LD document includes a context URL in its
```

lib/suites/LinkedDataSignature.js
```diff
@@ -311,8 +311,51 @@ module.exports = class LinkedDataSignature extends LinkedDataProof {
...
+ * `@context` property.
+ *
+ * @param {object} options - Options hashmap.
+ * @param {object} options.document - A JSON-LD document.
+ * @param {string} options.contextUrl - A context url.
```
mattcollier, line 349:
```suggestion
 * @param {string} options.contextUrl - A context URL.
```

mattcollier (approved):
TY.  Minor doc nits.

lib/suites/LinkedDataSignature.js
dmitrizagidulin: Thanks! fixing..

lib/suites/LinkedDataSignature.js
dmitrizagidulin: Fixed.

PR #149, opened 2021-07-01
Support multiple proof purposes in `verify`.

dlongley:
Includes cache of document c14n hash to avoid duplicate work when verifying multiple proofs against the same document.

### DIFF
--- a/lib/ProofSet.js
+++ b/lib/ProofSet.js
@@ -320,7 +320,10 @@ async function _matchProofSet({
       continue;
     }
 
-    // next, find the suite that can verify the proof
+    // next, find the suite that can verify the proof; if found, `matched`
+    // will be set to `true` and the proof will be added to `purposeToProofs`
+    // and `proofToSuite` to be processed -- otherwise it will not be; if
+    // no proofs are added for a given purpose, an exception will be thrown
     let matched = false;
     for(const s of suites) {
       // `matchingProofs` is a map of promises that resolve to whether a
--- a/lib/ProofSet.js
+++ b/lib/ProofSet.js
@@ -146,12 +146,14 @@ module.exports = class ProofSet {
 
       // verify proofs
       const results = await _verify({
-        document, suites, proofSet, purpose, documentLoader, expansionMap
+        document, suites, proofSet, purpose, documentLoader, expansionMap,
       });
       if(results.length === 0) {
-        throw new Error(
-          'Could not verify any proofs; no proofs matched the required ' +
-          'suite and purpose.');
+        const error = new Error(
+          'Did not verify any proofs; insufficient proofs matched the ' +
+          'acceptable suite(s) and required purpose(s).');
+        error.name = 'NotFoundError';
+        throw error;
       }
 
       // combine results
@@ -167,7 +169,7 @@ module.exports = class ProofSet {
       }
       return {verified, results};
     } catch(error) {
-      _addToJSON(error);
+      _makeSerializable(error);
       return {verified: false, error};
     }
   }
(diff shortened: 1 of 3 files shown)


dlongley:
This has been tested against zcapld and ed25519-signature-2020 using their separate test suites. We should really have a test suite here as well.

lib/ProofSet.js
```diff
@@ -146,12 +146,14 @@ module.exports = class ProofSet {
 
       // verify proofs
       const results = await _verify({
-        document, suites, proofSet, purpose, documentLoader, expansionMap
+        document, suites, proofSet, purpose, documentLoader, expansionMap,
```
msporny, line 149:
```suggestion
        document, suites, proofSet, purpose, documentLoader, expansionMap
```

Remove hanging comma.

lib/ProofSet.js
```diff
@@ -197,38 +199,108 @@ async function _getProofs({document}) {
...
+      };
+      const {verified, verificationMethod, error} = await suite.verifyProof({
+        proof, document, purpose, documentLoader, expansionMap
+      });
+      if(!vm) {
+        vm = verificationMethod;
       }
+      result = {proof, verified, verificationMethod: vm, error};
+    } catch(error) {
+      result = {proof, verified: false, error};
```
msporny, line 240: Do we need a `_makeSerializable(error)` here?

lib/ProofSet.js
```diff
@@ -197,38 +199,108 @@ async function _getProofs({document}) {
...
+        // validate purpose
+        const {verificationMethod} = result;
+        const suite = proofToSuite.get(proof);
+        let purposeResult;
+        try {
+          purposeResult = await purpose.validate(proof, {
+            document, suite, verificationMethod, documentLoader, expansionMap
+          });
+        } catch(error) {
+          purposeResult = {valid: false, error};
```
msporny, line 270: Again, do we need a `_makeSerializable(error)` call here?

lib/ProofSet.js
```diff
@@ -237,3 +309,52 @@ function _addToJSON(error) {
...
+  suiteMatchQueue, document, documentLoader, expansionMap
+}) {
+  for(const proof of proofSet) {
+    // first check if the proof matches the purpose; if it doesn't continue
+    if(!await purpose.match(proof, {document, documentLoader, expansionMap})) {
+      continue;
+    }
+
+    // next, find the suite that can verify the proof
+    let matched = false;
```
msporny, line 324: It's not clear what happens if `matched` continues to be `false`? I'm guessing `matches` is processed in some way in the calling function... hard to follow this code.

msporny (changes requested):
I'd like to see a test that tests the multiple proofs path, do we have one? If not, I suggest adding one before merging. I expect it would be easy for another developer to break multi-proof verification and not know they did without that test.

dlongley:
The multiple proofs test is over here:

https://github.com/digitalbazaar/zcapld/pull/64/files#diff-dce1aec60ad1f68ca409661764b88da658d166287e3e107738f901d0052be2e6R535

lib/ProofSet.js
dlongley: No, it happens below in one place: https://github.com/digitalbazaar/jsonld-signatures/pull/149/files#diff-01c3bd2f5d6eee0565c498c5535384249141782a41d6b8d5da84a6fc183a96e7R245

lib/ProofSet.js
dlongley: No, it happens below in one place: https://github.com/digitalbazaar/jsonld-signatures/pull/149/files#diff-01c3bd2f5d6eee0565c498c5535384249141782a41d6b8d5da84a6fc183a96e7R287

lib/ProofSet.js
dlongley: I'll add a comment about it.

dlongley: https://github.com/digitalbazaar/jsonld-signatures/pull/149/commits/bc539b6170398a2f2e7e791a9991b5648ef096ab

lib/ProofSet.js
dlongley: Removed.

msporny (approved):

Issue #150, opened 2021-07-08
Method `length` called on a Map


aljones15:
Looks like when some of these `_verify` methods were revised to use [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) from [`Array`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array) not all of the methods were updated:

https://github.com/digitalbazaar/jsonld-signatures/blob/10871df6e91421b20ddfae065baf688e7acdbfdc/lib/ProofSet.js#L201-L216

This code will consistently evaluate to

```js
undefined < purposes.length
// false
```

which will be false in almost all cases.

aljones15:
closing as issue has been resolved.

PR #151, opened 2021-07-08
Use size to get Map purposeToProofs entries length.

aljones15:
Addresses https://github.com/digitalbazaar/jsonld-signatures/issues/150

Corrects a minor issue where `length` was being called on a `Map` that needs `size`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # jsonld-signatures ChangeLog
 
+## 9.2.1 - TBD
+
+### Changed
+- Use the method size to get the number of entries in the Map purposeToProofs.
+
 ## 9.2.0 - 2021-07-02
 
 ### Added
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # jsonld-signatures ChangeLog
 
+## 9.2.1 - TBD
+
+### Changed
+- Use the method size to get the number of entries in the Map purposeToProofs.
```
mattcollier, line 6:
```suggestion
### Fixed
- Use the `size` method to get the number of entries in the `purposeToProofs` Map.
```

mattcollier (approved):
Good catch, would it be much trouble to add a test, or at least create an issue that describes a test that would have prevented this bug?  Thank you.

dlongley (approved):
Thanks, good catch!

dlongley:
I'll merge and release.

PR #153, opened 2021-07-11
Add optimization for controller docs that are DID docs.

dlongley:


dlongley:
This has been tested against zcapld and ed25519-signature-2020 repo test suites.

Issue #111, opened 2020-09-03
expansionMap cannot be overridden


JaceHensley:
According to the docs [here](https://github.com/digitalbazaar/jsonld-signatures/blob/master/lib/ProofSet.js#L41) you can set `expansionMap` to false to drop unmapped fields or provide custom mapper. But the implementation has `expansionMap` always being set to `strictExpansionMap` unless `expansionMap` is false. So only no expansion map OR the strict expansion map is allowed.

I'm happy to make a PR to update this to allow passing a custom expansionMap.

DenisPopov15:
Do we have any updates here? Looks like a bug here.

Issue #146, opened 2021-06-16
Add test cases that use new expansionMap capabilities


dlongley:
We need new test cases for at least:

1. Failing to sign (catch expansionMap error): A doc with a relative IRI `@id` value.
2. Failing to sign (catch expansionMap error): A doc with a relative IRI `@type` value w/no base IRI set.
3. Failing to sign (catch expansionMap error): A doc with a relative IRI `@type` value w/ base IRI set.

And the same 3 for failing to verify.

davidlehn:
Where do tests go now?  This repo has no tests anymore.  Adding to all the suites would be redundant.

dlongley:
@dmitrizagidulin, can you be of assistance here (where do the tests go)?

davidlehn:
While fixing a [related bug](https://github.com/digitalbazaar/jsonld-signatures/issues/111) it's still unclear where tests should be added.

dlongley:
Closing this issue as there are tests in cryptosuites for this -- just like there are tests in cryptosuites for other behavior. This issue is no longer special ... we still need to add a test suite here to enable better regression testing without having to run against some external cryptosuite tests.

PR #158, opened 2021-12-14
Allow `expansionMap` overrides.

davidlehn:
Fixes https://github.com/digitalbazaar/jsonld-signatures/issues/111.

Unsure where to add tests, (see https://github.com/digitalbazaar/jsonld-signatures/issues/146).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # jsonld-signatures ChangeLog
 
+## 9.3.1 - 2021-xx-xx
+
+### Fixed
+- Allow `expansionMap` overrides.
+
 ## 9.3.0 - 2021-07-10
 
 ### Added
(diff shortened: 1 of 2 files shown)


dlongley (approved):

davidlehn:
@dlongley I had bad logic in there the first pass. This is better. It looks like lower level handles `false` but maybe converting to `undefined` is better.

dlongley (approved):
Clearly a test here would be ideal as you note. Maybe the test could involve using some data with an undefined term and an expansion map that sets a flag and always throws -- then assert the flag has been set. Something simple and quick in the local test suite here.

davidlehn:
@dlongley Can non-trivial tests even be written in this module at the moment?  Previously the suites were here that ran the tests.  Probably some mock code is needed just to even make tests possible.  The code is indirectly tested through the suites, but that's not ideal to test common features.

dmitrizagidulin:
@davidlehn @dlongley in this particular PR, the expansionMap logic in `add()` and `verify()` could be tested via unit tests, by sinon-mocking the `suite` param.

PR #166, opened 2022-08-16
Use `jsonld@8` and `safe` mode instead of `expansionMap`.

dlongley:
We need some better tests here that exercise `safe` mode.


dlongley:
This has been tested against: https://github.com/digitalbazaar/ed25519-signature-2020

davidlehn:
- `jsonld.js` also throws if `expansionMap` is used.  Could that be used instead of all the checks here?  Or would that be too confusing where the error came from?
- I assume `lib/expansionMap.js` can be removed?
- It's still an [open question](https://github.com/digitalbazaar/jsonld-signatures/issues/146) of where/how to add tests for this base library.

davidlehn (approved):

PR #170, opened 2023-04-14
Update `matchProof()` to check if the `proof.cryptosuite` matches `suite.cryptosuite` along with   proof type for `DataIntegrityProof`.

JSAssassin:


dlongley (changes requested):
So I don't think we should change this here -- we should make sure that `DataIntegrityProof` overrides this method and performs the check over here: https://github.com/digitalbazaar/data-integrity/issues/4

JSAssassin:
Closing.

PR #172, opened 2023-05-13
Add `derive` API and pass existing `proofSet` to suite functions.

dlongley:


dlongley:
Addresses #171 and possibly others related to derived proofs.

